PR #26, opened 2020-06-16
don't load route planner key

qqnluaq:


qqnluaq:
The route planner key is not in the code, but can be set via configuration. For example:

https://bcgov.github.io/smk-client/debug/app/translink/?config={%22tools%22:[{%22type%22:%22directions%22,%22routePlannerService%22:{%22apiKey%22:%22[key]%22}}]}

Replace `[key]` with the route planner key

Use `https://qqnluaq.github.io/smk-client/` in the url if the above doesn't work.

PR #114, opened 2021-08-16
Updated esri-leaflet-legend-compat-src library

michaelpnelson:
This updates the file with latest changes from
https://github.com/w8r/esri-leaflet-legend/blob/master/dist/esri-leaflet-legend-compat-src.js.


dgboss:
I think we can abandon this PR. In general I don't like updating directly from a file in a repo and would rather update from a released package. Since there is no release later than the current 2.0.1 version (npm package, not tagged in GitHub), I don't think we need to take any action, unless there are important fixes we would be missing by no updating from the repo?

michaelpnelson:
I'm not aware of any fixes (or even advantages) of the changes in this commit, so I'm fine with abandoning this pull request.

PR #115, opened 2021-08-16
Updated leaflet library to version 1.7.1

michaelpnelson:
The new file is minified to improve performance.


dgboss:
As part of this PR, could you please remove the old versions (actual file and references in smk-tags.js) of leaflet, esri-leaflet, esri-leaflet-renderers?

michaelpnelson:
> As part of this PR, could you please remove the old versions (actual file and references in smk-tags.js) of leaflet, esri-leaflet, esri-leaflet-renderers?

I'm happy to make this change, but I'd prefer to do it in a separate pull request instead of adding it to this one - is that OK? It would make a potential rollback easier.

dgboss:
> > As part of this PR, could you please remove the old versions (actual file and references in smk-tags.js) of leaflet, esri-leaflet, esri-leaflet-renderers?
> 
> I'm happy to make this change, but I'd prefer to do it in a separate pull request instead of adding it to this one - is that OK? It would make a potential rollback easier.

Sure. that's fine. Please create a new ticket in JIRA for the additional work then. Also, feel free to cleanup any other extraneous files and references as you see fit. Thanks!

PR #116, opened 2021-08-18
Remove old Leaflet, Vue libs

michaelpnelson:
As libraries have been upgraded in smk, prior versions were commented out in smk-tags.js, and the library files left in /src/lib.

This change removes comments in smk-tags.js referencing prior libraries and removes prior library files from /src/lib/*.


dgboss (approved):
Looks good.

PR #118, opened 2021-09-09
showCoverageOnHover can be read from a vector layer

michaelpnelson:
This change allows smk to read a "showCoverageOnHover" value from a vector layer's "layer" element in smk-config.json. This change is a companion to a change in smk-cli that will add a UI control to turn the functionality on or off for a vector layer (https://github.com/bcgov/smk-cli/issues/63).

showCoverageOnHover can currently be read from the "viewer" element of smk-config.json. This allows a true value in the viewer to override a value in the layer.

If no value is set in a vector layer or in the viewer, the default value of 'false' is used.

###Suggested Test Procedure 

- Create/use an app using smk-cli
- Add a vector layer having points (Example:
https://openmaps.gov.bc.ca/geo/pub/WHSE_IMAGERY_AND_BASE_MAPS.GSR_ASSISTED_LIVING_LOCS_SV/ows?service=WMS&request=GetMap&version=1.1.1&layers=pub:WHSE_IMAGERY_AND_BASE_MAPS.GSR_ASSISTED_LIVING_LOCS_SV&format=application/json;type=geojson&transparent=true&cql_filter=include&srs=EPSG%3A4326&width=645&height=728&bbox=-144.19403793672322,45.39637993438814,-108.99384262912152,62.632414448856416)
- From smk:
-- check out these changes
-- run 'npm start' to copy changes from /src to /dist
-- run 'npm link' to link these changes to the global node modules
- From the directory of the app built with smk-cli:
-- run 'npm run view' to start the app and launch it in a browser
- Edit the smk app's smk-config.json and test different values for showCoverageOnHover in the "layer" element that has the vector layer as well as the "viewer" element
snippet:
"clusterOption": {
  "showCoverageOnHover": true
}
- Confirm that a hover over a cluster behaves as expected
- Test true, false, and no value (clusterOption, showCoverageOnHover elements removed)
- showCoverageOnHover should be false by default
- If a 'true' value for showCoverageOnHover is set in the viewer, it should be used
- If a value for showCoverageOnHover is set in the layer, it should be used and override any value in the viewer

### DIFF
--- a/docs/configuration/layers/cluster-option.md
+++ b/docs/configuration/layers/cluster-option.md
@@ -1,0 +1,20 @@
+###### [SMK](../../..) / [Configuration](..) / [Layers](.)
+
+# Layer Cluster Option Objects
+
+This object defines properties that influence the rendering of clusters of points.
+
+A clusterOption object contains these properties, which have these default values. 
+Click on a property name for more information:
+<pre>
+{
+    <a href="#showcoverageonhover-property"          >"showCoverageOnHover"</a>:          false
+}
+</pre>
+
+
+## ShowCoverageOnHover Property
+`"showCoverageOnHover": Boolean`
+
+If `true`, then the bounds of a cluster's markers show when you hover over the cluster.
+The default value is `false`.
--- a/docs/configuration/layers/style.md
+++ b/docs/configuration/layers/style.md
@@ -8,7 +8,7 @@ When multiple styles are defined, then each feature in the layer is rendered in 
 
 If a point feature is being rendered, the `"marker..."` properties are used. If the [`"markerUrl"` property](#markerurl-property) isn't defined, then a circle is rendered instead, using the `"stroke..."` and `"fill..."` properties. The circle radius will be half of the [`"strokeWidth"` property](#strokewidth-property).
 
-A style object contains is defined with these properties, which have these default values.
+A style object contains these properties, which have these default values. 
 Click on a property name for more information:
 <pre>
 {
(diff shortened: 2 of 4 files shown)


dgboss (changes requested):
I think that the layer configuration, if present, should be the ultimate source of truth. ie. A value configured at the viewer level should not override a value configured at the layer level. I think I would switch the order of the two if blocks on line 314 and 322. This would still maintain backwards compatibility for users who have enabled the option in an existing app, but then allow them to override individual layers using the new UI you are developing for the smk-cli. Thoughts?

michaelpnelson:
I had the misunderstanding that we wanted the viewer value (if true) to take precedence over a layer value. I'm happy to change this so that layer values take precedence.

docs/configuration/layers/cluster-option.md
```diff
@@ -0,0 +1,20 @@
+###### [SMK](../../..) / [Configuration](..) / [Layers](.)
+
+# Layer Cluster Option Objects
+
+This object defines properties that influence the rendering of clusters of points.
+
+A clusterOption object contains these properties, which have these default values.
```
dgboss, line 7: Maybe follow some of the existing pattern for other objects? eg. https://bcgov.github.io/smk/docs/configuration/layers/vector

"This is the default configuration for the clusterOption object."

michaelpnelson: I think the pattern "This is the default configuration for ..." is applied for docs on layers. Besides "vector", "wms" and "esri-dynamic" follow it, for example, "This is default configuration for the WMS layer".

I was following the pattern used by "style" and "attribute": "This object defines ...". Since clusterOption is a new object, I think this pattern is appropriate. Let me know if you disagree.

dgboss: I see. The language just sounds awkward to me. I prefer more 'natural language' in documentation. Change if you want or leave it as is, I'll leave it up to you.

dgboss (approved):

PR #123, opened 2021-09-17
Incrementing version to 1.0.6 in package.json

michaelpnelson:


dgboss (approved):

bdolor:
@michaelpnelson @dgboss I'm late to the party, though I wonder if this is was a potential opportunity to also update `package-lock.json`?  Though there will be no functional consequences with this particular PR, the habit of updating `package.json` with `package-lock.json` together, is likely a good one to practice. Currently there are different version numbers for `package.json` and `package-lock.json`.  

image

michaelpnelson:
Thanks @bdolor! I know to update package-lock when dependencies change, but it didn't occur to me that it might need updating in this situation. I'll follow up.

PR #125, opened 2021-09-23
Geomark tool added; can draw polygons to layer

michaelpnelson:
This adds support for a new "Geomark" tool (which will eventually be enabled/disabled in smk-cli). Its implementation has been modeled on that of the Version tool.

This also adds the ability to draw polygons and add them to a layer. This is implemented using the Leaflet Draw library, with new files copied from the latest version:  https://github.com/Leaflet/Leaflet.draw/archive/refs/tags/v0.4.14.zip.

To enable and test these changes in SMK, add this snippet to the "tools" section of smk-config.json:

        {
            "title": "Geomark",
            "position": "list-menu",
            "icon": "build",
            "type": "geomark",
            "enabled": true
        },


src/smk/tool/geomark/panel-geomark.html
```diff
@@ -0,0 +1,23 @@
...
+>
+    <template slot="header">
+        <slot></slot>
+    </template>
+
+    <h1>Geomark</h1>
+
+    <p><i>(This is a work in progress; functionality is yet to be added.)</i></p>
+
+    <p>Create a Geomark by drawing your an area on the map.</p>
```
dgboss: Can remove 'your'

dgboss:
A couple thoughts on the implementation so far:
1. I think it would be helpful if the draw tool became 'active' as soon as you click on the 'Geomark' menu item, so you can start clicking on the map immeditately, similar to the Geomark implementation on the firewood cutting DMF site (https://apps.gov.bc.ca/pub/dmf-viewer/?siteid=5342025102195394536).
2. In conjunction with the first point, is it possible to take advantage of the leaflet draw functionality without displaying the draw tool on the map?

dgboss:
One other comment, we already have a library included in the project for drawing markup on the map. I'm wondering if we could leverage this library instead of making use of a separate library? Check out src\smk\viewer-leaflet\tool\markup\lib and you will see the leaflet-pm library.

michaelpnelson:
> One other comment, we already have a library included in the project for drawing markup on the map. I'm wondering if we could leverage this library instead of making use of a separate library? Check out src\smk\viewer-leaflet\tool\markup\lib and you will see the leaflet-pm library.

Thanks @dgboss . 

Regarding having the polygon drawing start immediately when entering the Geomark tool: I think we should discuss this further. We could follow the DMF2 functionality more closely by having all functionality in the tool's window on the left and omit the tool bar on the right entirely. Or we could enable more of the tooling of the Leaflet Draw library and keep the tool bar on the right and some or all of its functionality. I started with the latter for simplicity, and because it offers useful functionality - for example, we can clear polygons individually instead of all at once; we can edit drawn polygons; there's Cancel functionality; and more. We could also consider enabling other shapes (circles, rectangles). I think it would be good to get feedback from @NicoledeGreef as well. I'm aiming for basic functionality first with the thinking that we can tune it in subsequent pull requests. It's also worth noting that the libraries used to draw polygons are different between DMF2 and this implementation (Google Maps vs. Leaflet Draw) and they have different capabilities, so duplicating DMF2's functionality precisely might require some shoehorning.

Regarding reusing existing library tooling: The Leaflet library code used to draw polygons and other shapes for the Markup tool only adds path elements to an SVG element, for visually highlighting areas. It doesn't keep lat/long coordinates for the points. Leaflet Draw does store layers and their coordinates from the drawn shapes.

michaelpnelson:
In the latest commit, the Leaflet Draw library is removed and the Geoman Leaflet Free library (https://github.com/geoman-io/leaflet-geoman) is added. 

Unfortunately I'm having trouble getting the Geoman library initialized. This error is caught in tools.js on loading SMK:
"Cannot read properties of undefined (reading 'pm')"

I'm not sure if we should/how we should initialize ... Leaflet Draw didn't seem to need it. Geoman's docs mention:

"Init Leaflet-Geoman
Just include leaflet-geoman.min.js right after Leaflet. It initializes itself."

michaelpnelson:
The "Updates" commit fixes the issue that prevented Geoman from loading. It also changes the way the tool is used - instead of adding controls, "draw mode" is entered when choosing the Geomark tool and left when leaving the Geomark tool. Options for draw mode are documented in https://github.com/geoman-io/leaflet-geoman#draw-mode.

dgboss:
It looks like there is some sort of conflict with the Identify tool. When I add the config for the Geomark tool to an existing app that has the Identify tool configured, I see the Geomark tool activate when I open the panel, but the tool deactivates as soon as I click on the map. Should this use case be supported yet?

michaelpnelson:
> It looks like there is some sort of conflict with the Identify tool. When I add the config for the Geomark tool to an existing app that has the Identify tool configured, I see the Geomark tool activate when I open the panel, but the tool deactivates as soon as I click on the map. Should this use case be supported yet?

I noticed something similar with the Location tool earlier - clicks to draw polygons would also invoke the location. I went looking for code that might filter click handling in the Markup tool, but it looks like that conflict isn't resolved - if both the Location and Markup tools are loaded, the same click will be handled by both. I think this is a bigger issue - Tools can step on each other this way. 

I'm glad you did testing with other Tools. I noticed while reacquainting myself with the Markup and Location issue that the Markup tool's toolbar doesn't display at all when the Markup tool is enabled in this branch, even when the Geomark tool is disabled - I'll have to look into that, as well as the issue with the Identify tool you mention.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -0,0 +1,49 @@
...
+                    smk.$viewer.map.pm.enableDraw('Polygon', {
+                        continueDrawing: true
+                    });
+                }
+                else {
+                    smk.$viewer.map.pm.disableDraw();
+                }
+            } )
+
+            var editableLayers = new L.FeatureGroup();
```
dgboss: I'm being a little pedantic, but could we use a more descriptive name here, like geomarkLayers?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -0,0 +1,49 @@
...
+            smk.$viewer.map.addLayer(editableLayers);
+
+            smk.$viewer.map.on('pm:create', function(e) {
+                var type = e.layerType,
+                    layer = e.layer;
```
dgboss: For the sake of readability, could you please declare each variable individually?
ie.
var type = e.layerType;
var layer = e.layer;

dgboss (changes requested):

src/smk/tool/geomark/panel-geomark.html
michaelpnelson: Done. This was copied from the Firewood Cutting Prohibited app, errors and all :)

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Done

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Done

dgboss (approved):

PR #126, opened 2021-09-29
Can create a geomark from drawn polygons

michaelpnelson:
This implements the basic functionality. I was hoping to use GeoJson format - Leaflet's LayerGroup has a neat "toGeoJson()" method - but the Geomarks API doesn't seem to accept the values, so I'm using WKT instead for now.

This may also need to change if we start drawing polygons using Geoman instead of Leaflet Draw.

I'd like to look into breaking the code into smaller functions. I'd also like to eventually add tests for the code that builds the string of coordinates, though of course we need to choose a testing framework first.


michaelpnelson:
I'll be creating a new pull request for this functionality. It will use a different branch rebased on earlier changes.

PR #129, opened 2021-10-06
Can create a Geomark

michaelpnelson:
This adds support for creating a Geomark from polygon coordinates in WKT format using the Geomark web service.

### DIFF
--- a/src/smk-tags.js
+++ b/src/smk-tags.js
@@ -13,6 +13,7 @@ t.script( 'document-ready', 'smk/document-ready.js' )
 t.script( 'jquery', 'lib/jquery-3.3.1.min.js' )
 t.script( 'vue', 'lib/vue-2.5.11.min.js' )
 t.script( 'turf', 'lib/turf-5.1.6.min.js' )
+t.script( 'geomark', 'https://apps.gov.bc.ca/pub/geomark/js/geomark.js', { external: true }  )
 
 t.script( 'libs', 'smk/libs.js' )
 // t.sequence( 'libs' )
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -21,9 +21,16 @@ include.module( 'tool-geomark', [
         function () {
             SMK.TYPE.ToolWidget.call( this, 'geomark-widget' )
             SMK.TYPE.ToolPanel.call( this, 'geomark-panel' )
+
+            this.defineProp( 'geomarkService' )
         },
         function ( smk ) {
             var self = this
+
+            if (!self.geomarkService) {
+                self.showStatusMessage('No value for "geomarkService" was found in configuration. Geomark tool functionality is disabled.', 'error', 5000);
+                return;
+            }
 
             this.buildLngLatCoords = function(layerGroup) {
                 var lngLatCoords = '';
@@ -68,13 +75,12 @@ include.module( 'tool-geomark', [
                 geomarkLayers.addLayer(layer);
             });
 
-            var baseUrl = 'https://apps.gov.bc.ca/pub/geomark';
-            var client = new window.GeomarkClient(baseUrl);
+            var client = new window.GeomarkClient(self.geomarkService.url);
 
             smk.on( this.id, {
                 'create-geomark': function () {
                     if (geomarkLayers.getLayers().length == 0) {
-                        alert('No drawings were found.');
+                        self.showStatusMessage('No drawings were found. Draw one or more polygons before creating a geomark.', 'warning');
                         return;
                     }
                     var lngLatCoords = self.buildLngLatCoords(geomarkLayers);
@@ -84,9 +90,10 @@ include.module( 'tool-geomark', [
                         'callback': function(geomarkInfo) {
                             var geomarkId = geomarkInfo.id;
                             if (geomarkId) { 
-                                alert('Created geomark: ' + geomarkInfo.url);
+                                alert('Created geomark: ' + geomarkInfo.url + 
+                                '. Save this URL to access your geomark later.');
                             } else {
-                                alert('Error creating geomark: ' + geomarkInfo.error);
+                                self.showStatusMessage('Error creating geomark: ' + geomarkInfo.error, 'error', 5000);
                             }
                         }
                     });
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -1,4 +1,5 @@
 include.module( 'tool-geomark', [ 
+    'geomark',
     'tool.tool-base-js',
     'tool.tool-widget-js',
     'tool.tool-panel-js',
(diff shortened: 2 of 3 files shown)


src/smk/tool/geomark/tool-geomark.js
```diff
@@ -44,6 +67,31 @@ include.module( 'tool-geomark', [
 
                 geomarkLayers.addLayer(layer);
             });
+
+            var baseUrl = 'https://apps.gov.bc.ca/pub/geomark';
+            var client = new window.GeomarkClient(baseUrl);
```
michaelpnelson: I think the baseUrl value should be inlined to line 72 ... I forgot to do this before pushing.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -44,6 +67,31 @@ include.module( 'tool-geomark', [
...
+
+            smk.on( this.id, {
+                'create-geomark': function () {
+                    if (geomarkLayers.getLayers().length == 0) {
+                        alert('No drawings were found.');
```
michaelpnelson: I use alerts here and below. Is there a favoured alternative to alerts we should use?

src/smk/tool/geomark/tool-geomark.js
dgboss: Could we read the baseUrl from configuration instead? I'm not crazy about hard coded URLs whenever possible. It would be a little more 'future proof' in the event that the URL were to ever change for some reason.

src/smk-tags.js
```diff
@@ -13,6 +13,7 @@ t.script( 'document-ready', 'smk/document-ready.js' )
 t.script( 'jquery', 'lib/jquery-3.3.1.min.js' )
 t.script( 'vue', 'lib/vue-2.5.11.min.js' )
 t.script( 'turf', 'lib/turf-5.1.6.min.js' )
+t.script( 'geomark', 'https://apps.gov.bc.ca/pub/geomark/js/geomark.js', { external: true }  )
```
dgboss, line 16: Just curious, have you checked what happens in the event this resource isn't available? Will SMK still load? Or is the entire app broken?

src/smk/tool/geomark/tool-geomark.js
dgboss: There is something called a 'status-message' in the smk project. I don't really know much about it, but it might be worth taking a look at? Check out src/smk/status-message if you're interested.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I've been looking for examples of URLs in configuration, but it looks like hard-coding is the norm in this project. Do you know if we even have (non-build-related) configuration files?

dgboss: I think the URL could be read from geomark config in smk-config.json. A hard coded default would then exist in the smk-cli project which would be written to the smk-config.json when adding the tool to an app.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Thanks for this suggestion. "Status message" is used often, and alert() isn't used elsewhere in tools, so I'll change this to use a status message.

One advantage is that, unlike an alert that the user must manually dismiss, the status message withdraws after 2 seconds.

One minor drawback is that the status message panel isn't clickable, and the cursor is still a crosshair with clicks added to drawn polygons, so users won't be able to click on a link in the status message. However, I think this is OK because, after creation, the Geomark will be added to a list (coming in another pull request) and that will have a clickable link. But if we want to follow the functionality in DMF2 more closely, perhaps it's worth having an alert here, just for the confirmation after a Geomark is created - let me know what you think. For now I will plan to replace all alerts with status messages.

dgboss: Thanks for looking into the status message option. It sounds like neither the alert nor the status message are great options. You might want to consider a couple other options:
1. A modal dialog built into the geomark Vue component 
2. A new modal component that can be used by other components

Here's an example modal for reference: https://vuejs.org/v2/examples/modal.html

Both DMF and the IMF2 Geomark implementations provide an easy way for a user to copy a Geomark's URL after it is created, so we probably want to maintain this ability in the SMK implementation. What do you think?

src/smk-tags.js
michaelpnelson: When I add characters to the URL to make it invalid, the SMK loads, but the Geoserver tool doesn't load.

dgboss: Perfect, thanks for checking!

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I've added support for geomark config in smk-config.json. Here is a snippet:

            "geomarkService": {
                "url": "https://apps.gov.bc.ca/pub/geomark"
            }

The full tool config would now look like:

        {
            "title": "Geomark",
            "position": "list-menu",
            "icon": "build",
            "type": "geomark",
            "geomarkService": {
                "url": "https://apps.gov.bc.ca/pub/geomark"
            },
            "enabled": true
        },

I've made this an object instead of a single property after seeing this done elsewhere in SMK - for example, the Directions/Waypoints tool declares objects for "geocoderService" and "routePlannerService".

I've also added a check for its presence, and notification if it's missing, so we can fail more gracefully if it's missing in config.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I vote to keep using an alert to display the URL when a geomark is created, but use the built-in status message functionality for the other notifications added. I think a modal solution is important so the user has time to copy the URL without it disappearing.

About your first suggestion ("A modal dialog built into the geomark Vue component") - I see that this would solve the issue of the status message being unclickable due to being in "draw mode", but I think it could be confusing to have some messages coming from the top centre and others coming from within the panel.

I'm fine with your second suggestion ("A new modal component that can be used by other components") and another suggestion is to add one of several Vue dialog libraries available (I found one article giving a top ten list of Vue dialog libraries, so there must be many). But I think either is a pretty big bite so I vote to make that another feature/pull request so this isn't blocked.

Let me know what you think of the changes for these notifications. I've made choices on additional functionality provided by "status message", including warn/error styling and milliseconds of delay before disappearing.

src/smk/tool/geomark/tool-geomark.js
dgboss: Awesome, thank you!

src/smk/tool/geomark/tool-geomark.js
dgboss: I feel that use of a browser's alert dialog leads to poor user experience, it just isn't very attractive or inline with the look and feel of the overall UI of the SMK app, so I would be hesitant to proceed with it. Perhaps we should engage Nicole? Given that this is a fairly UI intensive feature, it might be worth consulting our chapter's UX Designer (Joshua Jones) on the feature as a whole?

michaelpnelson: I agree that an alert isn't ideal, and I agree with your suggestion of adding a reusable modal dialog component. I'm just suggesting to do the modal dialog in a separate pull request so it doesn't hold this one back - see the 3rd paragraph of my earlier comment. The modal dialog component would be part of the collection of pull requests for the Geomark feature, so alerts would be replaced by the time we merge to master.

michaelpnelson:
After a chat with @dgboss , he agreed to defer the modal dialog change to another pull request in this Geomark feature branch.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -20,10 +21,39 @@ include.module( 'tool-geomark', [
...
+            this.defineProp( 'geomarkService' )
         },
         function ( smk ) {
             var self = this
 
+            if (!self.geomarkService) {
+                self.showStatusMessage('No value for "geomarkService" was found in configuration. Geomark tool functionality is disabled.', 'error', 5000);
+                return;
+            }
+
```
dgboss, line 34: Could you add a code comment explaining what this block does and why? I think you described it in a comment in the pull request, but having a record in the code will be helpful in the future if additional work is needed in this area.

PR #130, opened 2021-10-08
Leaflet PM lib removed

michaelpnelson:
This removes the Leaflet PM library, used by the Markup tool. Its functionality now comes from the Leaflet-Geoman library, which was added in an earlier commit and is a drop-in replacement for Leaflet PM.

Options for adding UI controls in the toolbar have been updated. Many of these are the same as what was available in Leaflet PM, but some have also been added in Leaflet-Geoman. Options that set a default value have been removed. New options that were not available or enabled earlier are disabled. As a result, we have the same six controls in the toolbar as before.


michaelpnelson:
I did some testing using the Geomark tool with the Markup tool enabled at the same time. The toolbar controls added by Markup can be used to edit and remove polygons created by the Geomark tool. The one issue I noticed is that editing and removing polygons exits "draw mode", where the mouse is a cursor and clicks add polygon points. Exiting and reopening the Geomark panel restores draw mode. We might consider adding a note on the panel mentioning that the use of Markup toolbar controls can impact draw mode.

dgboss:
I played around with the markup tools and they work great. I did notice one change, with the old library, lines would be red when you were drawing a line or shape and then turn blue when you finished drawing.

With the new library, the lines are always blue. Is there a way to set the color of the lines while in draw mode?

michaelpnelson:
> I played around with the markup tools and they work great. I did notice one change, with the old library, lines would be red when you were drawing a line or shape and then turn blue when you finished drawing.
> 
> With the new library, the lines are always blue. Is there a way to set the color of the lines while in draw mode?

I spent some time looking for a way to change the line color while drawing (for only the Markup tool) but didn't find a good solution.

Geoman's options for drawing are at https://github.com/geoman-io/leaflet-geoman#draw-mode, and drawing style options are at https://github.com/geoman-io/leaflet-geoman#customize-style. I experimented by adding this code to tool-markup-leaflet.js:

```
        smk.$viewer.map.on('pm:drawstart', function(e) {
            console.log(e);
            smk.$viewer.map.pm.setPathOptions({
                color: 'red'
            });
        });

        smk.$viewer.map.on('pm:create', function(e) {
            console.log(e);
            smk.$viewer.map.pm.setPathOptions({
                color: '#3388ff'
            });
        });
```

(#3388ff is the blue we see when drawing.)

The path options set on an event don't seem to apply when I expect them to. In my testing of this code, lines are blue as I draw them, then become red when I close the polygon (and "pm:create" is triggered). 

I might be able to find the combination of events that will give the behaviour we want. However there's a bigger issue: because events are applied to the map, they are used everywhere, so it looks we can't set options just for the Markup tool.  

UPDATE: I also tried setting the "templineStyle" option, but that didn't seem to do what I expected:

```
        smk.$viewer.map.pm.setPathOptions(
            {
                templineStyle: { color: 'red' }
            }
        );
```

dgboss:
Maybe give something like this a try (with appropriate RGB colors of course!).

`        var globalOptions = {
            templineStyle: { color: 'purple' },
            hintlineStyle: { color: 'orange', dashArray: [15, 5] }
        };
        smk.$viewer.map.pm.setGlobalOptions(globalOptions);`

Out of scope for this PR, but something to keep in mind, you could also set a different color when drawing Geomarks.

michaelpnelson:
I haven't yet found a way to scope the style changes to just the Markup tool. "tool-markup-leaflet.js"'s code doesn't seem to catch changedActive() events (like the Geomark JS does), probably because it doesn't have a panel component. And catching "map" events like "pm:drawStart" and "pm:create" applies to all tools because there's only one map.

What do you think of deferring this as a potential future enhancement (to be revisited before merging to master), and going ahead with this PR for now?

dgboss:
> I haven't yet found a way to scope the style changes to just the Markup tool. "tool-markup-leaflet.js"'s code doesn't seem to catch changedActive() events (like the Geomark JS does), probably because it doesn't have a panel component. And catching "map" events like "pm:drawStart" and "pm:create" applies to all tools because there's only one map.
> 
> What do you think of deferring this as a potential future enhancement (to be revisited before merging to master), and going ahead with this PR for now?

Feel free to defer the work, but please file an issue in Jira so we don't lose sight of it.

michaelpnelson:
> Feel free to defer the work, but please file an issue in Jira so we don't lose sight of it.

I've created a JIRA task for this. If you have no further suggestions, please approve this change.

dgboss (approved):
Approved

PR #131, opened 2021-10-12
Geomarks are listed

michaelpnelson:
As geomarks are created, they are added to a list. Each geomark in the list can be toggled on and off.

### DIFF
--- a/src/smk/tool/geomark/panel-geomark.html
+++ b/src/smk/tool/geomark/panel-geomark.html
@@ -20,4 +20,15 @@
         <li>Load an Existing Geomark</li>
         <li>Create a Geomark from File</li>
     </ul>
+
+    <h2>Loaded Geomarks</h2>
+
+    <div id="geomarks-container">
+        <ul>
+            <li v-for="geomark in geomarks">
+                <input type="checkbox" checked v-bind:id="geomark.id" v-on:click="$$emit( 'toggle-geomark', { id: geomark.id } )">
+                <a v-bind:href="geomark.url">{{ geomark.id}}</a>
+            </li>
+        </ul>
+    </div>
 </tool-panel>
--- a/src/smk/tool/geomark/panel-geomark.html
+++ b/src/smk/tool/geomark/panel-geomark.html
@@ -21,11 +21,8 @@
         <li>Create a Geomark from File</li>
     </ul>
 
-    <h2>Loaded Geomarks</h2>
-
-    <p v-if="geomarks.length === 0">No geomarks have been loaded.</p>
-
-    <div id="geomarks-container">
+    <div v-if="geomarks.length != 0" id="geomarks-container">
+        <h2>Loaded Geomarks</h2>
         <ul>
             <li v-for="geomark in geomarks">
                 <input type="checkbox" checked v-bind:id="geomark.id" v-on:click="$$emit( 'toggle-geomark', { id: geomark.id } )">
--- a/src/smk/tool/geomark/panel-geomark.html
+++ b/src/smk/tool/geomark/panel-geomark.html
@@ -23,6 +23,8 @@
 
     <h2>Loaded Geomarks</h2>
 
+    <p v-if="geomarks.length === 0">No geomarks have been loaded.</p>
+
     <div id="geomarks-container">
         <ul>
             <li v-for="geomark in geomarks">
--- a/src/smk/tool/geomark/panel-geomark.html
+++ b/src/smk/tool/geomark/panel-geomark.html
@@ -23,8 +23,8 @@
 
     <div v-if="geomarks.length != 0" id="geomarks-container">
         <h2>Loaded Geomarks</h2>
-        <ul>
-            <li v-for="geomark in geomarks">
+        <ul class="smk-no-list-style">
+            <li class="smk-truncate-with-ellipsis" v-for="geomark in geomarks">
                 <input type="checkbox" checked v-bind:id="geomark.id" v-on:click="$$emit( 'toggle-geomark', { id: geomark.id } )">
                 <a v-bind:href="geomark.url">{{ geomark.id}}</a>
             </li>
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -13,7 +13,8 @@ include.module( 'tool-geomark', [
 
     Vue.component( 'geomark-panel', {
         extends: SMK.COMPONENT.ToolPanelBase,
-        template: inc[ 'tool-geomark.panel-geomark-html' ]
+        template: inc[ 'tool-geomark.panel-geomark-html' ],
+        props: [ 'geomarks' ]
     } )
     // _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
     //
@@ -23,6 +24,9 @@ include.module( 'tool-geomark', [
             SMK.TYPE.ToolPanel.call( this, 'geomark-panel' )
 
             this.defineProp( 'geomarkService' )
+            this.defineProp( 'geomarks' )
+
+            this.geomarks = []
         },
         function ( smk ) {
             var self = this
@@ -30,6 +34,12 @@ include.module( 'tool-geomark', [
             if (!self.geomarkService) {
                 self.showStatusMessage('No value for "geomarkService" was found in configuration. Geomark tool functionality is disabled.', 'error', 5000);
                 return;
+            }
+
+            this.createCurrentDrawingLayer = function() {
+                var drawingLayer = new L.FeatureGroup();
+                smk.$viewer.map.addLayer(drawingLayer);
+                return drawingLayer;
             }
 
             this.buildLngLatCoords = function(layerGroup) {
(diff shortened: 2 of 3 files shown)


src/smk/tool/geomark/panel-geomark.html
```diff
@@ -20,4 +20,15 @@ <h1>Geomark</h1>
         <li>Load an Existing Geomark</li>
         <li>Create a Geomark from File</li>
     </ul>
+
+    <h2>Loaded Geomarks</h2>
```
dgboss: Could we hide this when there are no Geomarks in the list?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -23,6 +24,9 @@ include.module( 'tool-geomark', [
             SMK.TYPE.ToolPanel.call( this, 'geomark-panel' )
 
             this.defineProp( 'geomarkService' )
+            this.defineProp( 'geomarks' )
```
dgboss: Could you please add a semi-colon at the end of these lines?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -65,25 +75,42 @@ include.module( 'tool-geomark', [
                 }
             } )
 
-            var geomarkLayers = new L.FeatureGroup();
-            smk.$viewer.map.addLayer(geomarkLayers);
+            var currentDrawingLayer = self.createCurrentDrawingLayer();
```
dgboss, line 78: Stylistic thing, but how do you feel about initializing all the variables in a given scope at the same location (eg move this up to line 33 and move 'var client' at line 105 up as well)? Versus sprinkling them throughout?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -65,25 +75,42 @@ include.module( 'tool-geomark', [
...
+                    url: geomarkInfo.url,
+                    drawingLayer: drawingLayer
+                }
+            }
+
+            this.getGeomarkById = function(geomarkId) {
+                var matchingGeomark;
+                self.geomarks.forEach(function(item, index, array) {
+                    if (item.id === geomarkId) {
+                        matchingGeomark = item;
```
dgboss: Can we either break out of this for loop when we find a match? Or return the item directly from the for loop?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -65,25 +75,42 @@ include.module( 'tool-geomark', [
...
+                currentDrawingLayer.addLayer(layer);
             });
 
             var client = new window.GeomarkClient(self.geomarkService.url);
 
             smk.on( this.id, {
                 'create-geomark': function () {
-                    if (geomarkLayers.getLayers().length == 0) {
+                    if (currentDrawingLayer.getLayers().length == 0) {
                         self.showStatusMessage('No drawings were found. Draw one or more polygons before creating a geomark.', 'warning');
```
dgboss: I didn't catch this in the last PR, but is it possible to increase the delay before the message is dismissed. I think it is a little on the short side right now.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -92,11 +119,22 @@ include.module( 'tool-geomark', [
...
+                                currentDrawingLayer = self.createCurrentDrawingLayer();
                             } else {
                                 self.showStatusMessage('Error creating geomark: ' + geomarkInfo.error, 'error', 5000);
                             }
                         }
                     });
+                },
+                'toggle-geomark': function(idObj) {
+                    var geomark = self.getGeomarkById(idObj.id);
+                    if (!geomark) return;
```
dgboss: Could you expand this out to
```
if (!geomark) {
    return;
}
```

dgboss (changes requested):

src/smk/tool/geomark/panel-geomark.html
michaelpnelson: Good suggestion! 

I looked at what the Firewood app does, and I will copy that functionality: when no geomarks have been created/loaded, it displays "No geomarks have been loaded.".

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Sure. I'm assuming we want to do end lines in semicolons in general? I've been doing this mostly but I'm sure I forget at times.

At some point we can consider adding a linter to this project to help us stay consistent on style.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: My biggest counter-argument is readability: for someone new to the code, the variable is defined closely to where it is used, whereas if all variables were defined in the same place near the top, the reader starting far down has to go way up to see the definition.

I prefer to keep functions small to begin with, so where I can I like to break long functions into smaller ones. This one is pretty long already and there's much more code to come :) .

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Thanks; this is inefficient.

I revisited the methods on Array, and I think Array.find() works well here - it looks like an even simpler solution - so I'll use that.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Sure, I'll increase it to 5 seconds (like the other usages of "showStatusMessage" here) from the default of 2 seconds.

src/smk/tool/geomark/tool-geomark.js

src/smk/tool/geomark/panel-geomark.html
dgboss: My suggestion would be to hide this portion of the UI until a Geomark is actually drawn/created or loaded. The current phrasing in the Firewood DMF app could be confusing if a user's intent is to draw a new Geomark and not 'load' an existing one.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -9,36 +9,46 @@ include.module( 'tool-geomark', [
...
                 return;
             }
 
+            this.createCurrentDrawingLayer = function() {
+                var drawingLayer = new L.FeatureGroup();
+                smk.$viewer.map.addLayer(drawingLayer);
+                return drawingLayer;
+            }
+
             this.buildLngLatCoords = function(layerGroup) {
```
dgboss, line 45: I've discovered that this function doesn't generate valid WKT geometries for multiple polygons. No need to fix as part of this PR, but could you please file an issue to address in the future? To reproduce:
1. Draw two polygons on the map
2. Click 'Save drawing as a Geomark
3. Copy the generated URL
4. Open the URL in a browser and look for the 'Is Valid Geometry' field which will have a value of false.

eg. https://apps.gov.bc.ca/pub/geomark/geomarks/gm-E7B27F0F92C844AEB632E67E0D15B5C1

src/smk/tool/geomark/tool-geomark.js
dgboss: Your choice!

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -65,25 +75,38 @@ include.module( 'tool-geomark', [
...
+            }
+
+            this.getGeomarkById = function(geomarkId) {
+                return self.geomarks.find(function(item) {
+                    return item.id === geomarkId;
+                });
+            }
 
             smk.$viewer.map.on('pm:create', function(e) {
                 var type = e.layerType;
```
dgboss: It doesn't look like this is used anywhere (var type = e.layerType;). I think you can delete it.

src/smk/tool/geomark/panel-geomark.html
michaelpnelson: Done.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I've observed this as well. My code doesn't work well with multipolygons. 

This will be addressed in a forthcoming pull request where the ability to load a geomark is added. Instead of parsing geometries myself I'll be using a library.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Good catch! Removed.

src/smk/tool/geomark/panel-geomark.html
```diff
@@ -20,4 +20,14 @@ <h1>Geomark</h1>
...
     </ul>
+
+    <div v-if="geomarks.length != 0" id="geomarks-container">
+        <h2>Loaded Geomarks</h2>
+        <ul>
+            <li v-for="geomark in geomarks">
+                <input type="checkbox" checked v-bind:id="geomark.id" v-on:click="$$emit( 'toggle-geomark', { id: geomark.id } )">
+                <a v-bind:href="geomark.url">{{ geomark.id}}</a>
+            </li>
+        </ul>
```
dgboss, line 31: I think there is room for some improvement of the UI. I don't think an unordered list needs to used to display the geomarks. I feel that checkboxes with a slight indentation would be sufficient. Thoughts?
 
Also, the name of the geomark ends up getting truncated. Can you come up with a way to either fit the entire geomark name within the panel, or perhaps truncate with ellipsis and show the full name on hover?

image

dgboss: The bullets are being removed and further enhancements to the UI are being deferred.

dgboss (approved):
.

src/smk/tool/geomark/panel-geomark.html
michaelpnelson: Done. In the next commit, the bullets are removed from the list items and extra space before items is removed; and item text is truncated with ellipses. ul and li tags remain to continue to identify this semantically as a list for non-visual processing.

PR #132, opened 2021-10-18
Can load a geomark from a URL

michaelpnelson:


src/smk/tool/geomark/tool-geomark.js
```diff
@@ -120,6 +136,31 @@ include.module( 'tool-geomark', [
...
                 },
+                'load-geomark': function() {
+                    var geomarkUrl = prompt('Enter the URL of a geomark to load:');
+                    var geomarkId = self.extractGeomarkId(geomarkUrl);
+                    if (!geomarkId) {
+                        self.showStatusMessage('Could not discern a geomark ID within "' + geomarkUrl + '"', 'warning', 5000);
+                        return;
+                    }
+                    if (self.getGeomarkById(geomarkId)) {
+                        self.showStatusMessage('Geomark ' + geomarkId + ' is already loaded.', 'warning');
```
dgboss: What do you think about extending the length of time the message is visible for to 5 seconds?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -120,6 +136,31 @@ include.module( 'tool-geomark', [
...
+                        return;
+                    }
+                    $.ajax({
+                        url: geomarkUrl + '/feature.geojson',
+                        dataType: 'json',
+                        cache: false,
```
dgboss: Is there a reason we don't want the browser caching the response?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -120,6 +136,31 @@ include.module( 'tool-geomark', [
...
+                        traditional: true,
+                        success: function(geomarkFeature) {
+                            var polygon = L.geoJSON(geomarkFeature.geometry).addTo(smk.$viewer.map);
```
dgboss: Will this always be a polygon? Or can we also load point and line geomarks? What do you think about naming this to help indicate that 'polygon' is actually a layer of some sort? I didn't realize it was a layer until later on when I saw it passed to another function and then I looked at that function's signature.

dgboss: Is the result of L.geoJson() guranteed to be defined? ie. Will we never see a undefined error when the .addTo() call is made?

In general, I'm not a huge fan of chaining functions together as it is easy to run into the situation I'm asking about.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -42,28 +46,40 @@ include.module( 'tool-geomark', [
...
+                            wktCoords += ', ';
                         }
                     });
                 });
-                return lngLatCoords;
+                return wktCoords + ')';
             }
 
+            this.extractGeomarkId = function(geomarkUrl){
+                if (!geomarkUrl) return;
```
dgboss: Could you expand this if statement please?
```
if (!geomarkUrl) {
  return;
}
```

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -42,28 +46,40 @@ include.module( 'tool-geomark', [
...
+                var trimmedUrl = geomarkUrl.trim();
+                var lastSlashIdx = trimmedUrl.lastIndexOf('/');
```
dgboss: A trivial readability thing, but I'd suggest just spelling out 'index' within the variable name.

dgboss: It's quite common for people to include a trailing slash on a URL. This line will cause the addition of a geomark to fail if the user includes a trailing slash, with no indication as to why the attempt failed. I would suggest handling this situation in some way.

dgboss (changes requested):
It's working great!

One request that I neglected to mention in the previous PR (listing geomarks), could you please have the hyperlinks for geomarks open in new tabs/windows instead of in the current tab? Thanks!

michaelpnelson:
> One request that I neglected to mention in the previous PR (listing geomarks), could you please have the hyperlinks for geomarks open in new tabs/windows instead of in the current tab? Thanks!

Done.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Of the messages, this one is quite short and quick to digest, so I thought the default of 2 seconds is adequate, but I can extend it.

We might consider increasing the 2 second default at some point.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: No reason - I just copied the settings used by the client (https://apps.gov.bc.ca/pub/geomark/js/geomark.js).

Given that a geomark is read-only, I don't see the value of preventing caching because we wouldn't expect the response to have changed data. I'll remove the "cache" parameter as it is true by default.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I will rename "polygon" because it could be a line or a point (I tested a line but the geomark service didn't accept my point so that may not be supported by geomarks).

About L.geoJson() ... I did some testing, and if I give it garbage data, nothing breaks in the app, the layer doesn't load, and the console logs an error: "Invalid GeoJSON object.". We could trap for it, but I'm not sure we need to be validating the response data, and should there ever be a problem with response data, logging the error seems a good way to expose it.

src/smk/tool/geomark/tool-geomark.js

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I've renamed the variable to include the full word "index".

Good point about the trailing slash. I've added validation code that will trim a trailing slash. I realized we should also attempt to prevent malicious data so I added other sanitation code to encode characters and remove query parameters. 

I considered adding validation that the geomark ID begin with "gm-", but I hesitate to in case there are legacy geomarks without this pattern, or if the pattern changes.

src/smk/tool/geomark/tool-geomark.js
dgboss: Sounds good

src/smk/tool/geomark/tool-geomark.js
dgboss: I'm a slow reader. Thanks for changing!

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -71,12 +71,23 @@ include.module( 'tool-geomark', [
                 return wktCoords + ')';
             }
 
+            this.tidyUrl = function(url) {
+                url = url.trim();
+                if (url.endsWith('/') && url.length > 0) {
```
dgboss: I don't think this condition is necessary. If an URL ends with a '/' character, it will always have length greater than 0.

src/smk/tool/geomark/tool-geomark.js
dgboss: Excellent!

dgboss (changes requested):
One more small tweak and it's good to go!

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Good point!

dgboss (approved):
.

PR #133, opened 2021-10-19
Can clear current drawing

michaelpnelson:
This adds the ability to clear layers from the current drawing layer.

### DIFF
--- a/src/smk/tool/geomark/panel-geomark.html
+++ b/src/smk/tool/geomark/panel-geomark.html
@@ -17,6 +17,7 @@
 
     <ul>
         <li><a href="#" v-on:click="$$emit( 'create-geomark' )">Save Drawing as Geomark</a></li>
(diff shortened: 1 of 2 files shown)


dgboss (approved):
Just a random thought for you consideration. Should the clear drawing option delete all the polygons you have added to the map, but not yet saved as a Geomark? Or only the last drawing you added to the map? Personally I'd prefer one polygon be deleted per action, but leave it up to you.

michaelpnelson:
> Just a random thought for you consideration. Should the clear drawing option delete all the polygons you have added to the map, but not yet saved as a Geomark? Or only the last drawing you added to the map? Personally I'd prefer one polygon be deleted per action, but leave it up to you.

The current implementation (clearing all unsaved polygons) is the same behaviour as in the reference Firewood app. We can reconsider this when we meet to review usability, but I'll go ahead with this for now. 

I'll address the merge conflicts separately.

PR #134, opened 2021-10-19
Can create a geomark from a file

michaelpnelson:


michaelpnelson:
This change looks fine in Chrome and Firefox but is awkward in Safari. On clicking the link, a new tab opens but is empty; if you go back to the original tab, the alert is waiting. Once the alert is cleared, the Geomark "create from file" page is loaded in the new tab.

I'll see if I can find a workaround for this, but since it's pretty straightforward Javascript running (a window.open() followed by an alert(), nothing fancy), I'm hesitant to have a workaround that impacts other browsers.

UPDATE: This awkward behaviour can be seen in Safari by opening a file with this HTML content ...

```
<html>
    <head>
        <script>
            function openTabAndAlert() {
                window.open('https://apps.gov.bc.ca/pub/geomark/geomarks#file');
                alert('Hey, an alert!');
            }
        </script>
    </head>
    <body>
        <a onclick="openTabAndAlert()" href="#">Open link in new tab and show alert!</a>
    </body>
</html>
```

dgboss (changes requested):
I don't think you need a Safari specific workaround or hack. You should be able to fix the problem by simply opening the alert before opening the new tab/window. alert() is a blocking call, so window.open() will not be invoked until the user closes the alert.   This gives you an opportunity to explain the actions the user must take before we open a new tab that might hide the alert.

michaelpnelson:
> I don't think you need a Safari specific workaround or hack. You should be able to fix the problem by simply opening the alert before opening the new tab/window. alert() is a blocking call, so window.open() will not be invoked until the user closes the alert. This gives you an opportunity to explain the actions the user must take before we open a new tab that might hide the alert.

Thanks. I'd tried swapping the lines earlier and I recall seeing issues. But trying it now locally, it works as intended for all browsers, so I must have had some other complication earlier.

I'll swap the lines.

dgboss (approved):
Looks good

PR #136, opened 2021-10-28
Add modal components

michaelpnelson:


src/smk/component/alert/component-alert.css
```diff
@@ -0,0 +1,33 @@
+.modal-mask {
```
michaelpnelson: These changes are mostly as-is from the Vue example.

src/smk/component/alert/component-alert.css
```diff
@@ -0,0 +1,33 @@
...
+    position: fixed;
+    z-index: 99999;
```
michaelpnelson: Even with this high z-index, the dialog box still appears underneath the "smk toolbar" div when the window is narrow. I'm looking into a fix but if you know of one please let me know.

src/smk/component/alert/component-alert.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.modal-wrapper {
+    display: table-cell;
+    top: 20px;
+}
+
+.modal-container {
+    font-size: 12px;
+    width: 500px;
```
michaelpnelson: Fixed width, or make this a variable? I’d started coding this as a variable (so one could specify a wider width so that geomark URLs could display fully), though I didn’t get it fully working. I also realized different modal widths could be confusing to the user.

src/smk/component/alert/component-alert.css
```diff
@@ -0,0 +1,33 @@
+.modal-mask {
+    position: fixed;
```
michaelpnelson: I’m planning to pull out common CSS to another file. My vote is component.css - what do you think?

src/smk/component/alert/component-alert.css
```diff
@@ -0,0 +1,33 @@
...
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background-color: rgba(0, 0, 0, 0.5);
+    display: table;
+}
+
+.modal-wrapper {
+    display: table-cell;
+    top: 20px;
```
michaelpnelson: “Top” is ignored because this box is not positioned. I’ll look into another way drop it. As a counter argument, some OS’s have their alerts at the very top.

src/smk/component/prompt/component-prompt.html
```diff
@@ -0,0 +1,18 @@
...
+      <div class="modal-body">
+        <slot name="promptBody"></slot>
+      </div>
+      <div class="modal-body">
+        <input type="text" v-bind:value="value" @input="$emit('input', $event.target.value)">
+      </div>
+      <div class="modal-footer">
+        &nbsp;
+        <button class="modal-default-button" @click="$emit('cancel')">Cancel</button>
+        &nbsp;
```
michaelpnelson: This doesn't seem to result in adequate space between the two buttons. I'm looking into other ways but let me know if you have any suggestions.

src/smk/tool/geomark/panel-geomark.html
```diff
@@ -29,4 +29,16 @@ <h2>Loaded Geomarks</h2>
             </li>
         </ul>
     </div>
+
+    <alert v-if="showAlert === true" @close="$$emit( 'close-alert' )">
```
michaelpnelson: Since there is only one component, we have to change component properties before showing it. Alternately we could have multiple alert components for this panel, one for each alert usage we need, but that seems heavier.

src/smk/component/alert/component-alert.css
dgboss: Yes, I think that is a good idea.

src/smk/component/alert/component-alert.css
dgboss: Do you have a screenshot of what you mean?

src/smk/component/alert/component-alert.css
dgboss: Generally speaking, it is pretty common for web apps to display modal at the center of the screen as they require action from the user to dismiss. I'd suggested we display the alert and prompt components at the center of the screen.

src/smk/component/alert/component-alert.css
dgboss: A fixed width might be ok, but you would also need to set a max-width so the modal doesn't get cut off on small, mobile screens. Maybe max-width: 90%?

src/smk/component/alert/component-alert.html
```diff
@@ -0,0 +1,13 @@
+<div class="modal-mask">
+  <div class="modal-wrapper">
+    <div class="modal-container">
+      <div class="modal-body">
+        <slot name="alertBody"></slot>
+      </div>
+      <div class="modal-footer">
+        &nbsp;
+        <button class="modal-default-button" @click="$emit('close')">OK</button>
```
dgboss: In order to avoid potential collisions among handlers, it would be good to make the event name less 'generic'. I'd suggest 'alert-close' or something along those lines.

src/smk/component/prompt/component-prompt.html
```diff
@@ -0,0 +1,18 @@
...
+        <button class="modal-default-button" @click="$emit('close', value)">OK</button>
```
dgboss: Same thing here with event naming.

src/smk/component/alert/component-alert.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.modal-container {
+    font-size: 12px;
```
dgboss: I don't think we should reduce the font size. You can simply remove this line.

src/smk/component/alert/component-alert.css
dgboss: Also be aware that I am seeing css collisions due to identical naming of css classes in the alert and prompt components. Pulling out common css to a single file as you suggested will alleviate some of this. If all css will not be shared, it is good practice to 'scope' the css names. I'd suggest prefixing alert specific css class names with 'alert-'. Same thing for prompt specific css.

src/smk/component/alert/component-alert.js
```diff
@@ -0,0 +1,14 @@
...
+    'component',
+    'component-alert.component-alert-html' 
+], function ( inc ) {
+    "use strict";
+
+    Vue.component('alert', {
+        template: inc[ 'component-alert.component-alert-html' ],
+        data: {
+            alertBody: '',
+            showAlert: false
```
dgboss: Does showAlert get used anywhere?

src/smk/component/prompt/component-prompt.html
```diff
@@ -0,0 +1,18 @@
...
+    <div class="modal-container">
+      <div class="modal-body">
+        <slot name="promptBody"></slot>
+      </div>
+      <div class="modal-body">
+        <input type="text" v-bind:value="value" @input="$emit('input', $event.target.value)">
+      </div>
+      <div class="modal-footer">
+        &nbsp;
+        <button class="modal-default-button" @click="$emit('cancel')">Cancel</button>
```
dgboss: In order to avoid potential collisions among handlers, it would be good to make the event name less 'generic'. I'd suggest 'prompt-cancel' or something along those lines.

src/smk/component/prompt/component-prompt.html
```diff
@@ -0,0 +1,18 @@
+<div class="modal-mask">
+  <div class="modal-wrapper">
+    <div class="modal-container">
+      <div class="modal-body">
+        <slot name="promptBody"></slot>
+      </div>
+      <div class="modal-body">
+        <input type="text" v-bind:value="value" @input="$emit('input', $event.target.value)">
```
dgboss: This works just fine and you don't need to change it, I just wanted to mention that Vue.js has some syntactic sugar available to cover this type of binding. I believe this would accomplish the same thing:

`<input type="text" v-model="value">`

src/smk/component/prompt/component-prompt.html
dgboss: There are very few situations where you would actually want to use a non-breaking space. You can adjust your spacing using css. Your buttons have a css class in which you can use the css 'margin' property to add some spacing. For example, you could apply margin-left: 5px to the modal-default-button class. Or if you only want spacing on one button, wrap that button in a div, specify a new css class and add the left margin there.

src/smk/component/alert/component-alert.css
michaelpnelson: I reduced the font size so that it matches the size in the "show status" component. 

What do you think of enlarging the font size in "show status"?

src/smk/tool/geomark/panel-geomark.html
dgboss: Yep, this is totally fine since you can't have two alerts showing at once.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -102,6 +120,13 @@ include.module( 'tool-geomark', [
                 }
             } )
 
+            this.createAndShowAlert = function(alertBodyArg) {
+                self.alertBody = alertBodyArg;
+                self.showAlert = true;
+            }
+
+            this.handlePrompt = function(promptValue) {};
```
dgboss: I'm not quite clear what this is for?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -102,6 +120,13 @@ include.module( 'tool-geomark', [
                 }
             } )
 
+            this.createAndShowAlert = function(alertBodyArg) {
```
dgboss: Trivial, but maybe consider renaming this function. When I see 'create', I expect something new to be instnatiated. Perhaps just showAlert or updateAndSHowAlert?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -151,33 +205,13 @@ include.module( 'tool-geomark', [
                     });
                 },
                 'create-geomark-from-file': function () {
-                    alert('Upload your file using the form in the new window. Once you have a Geomark URL, load it using "Load an Existing Geomark".');
+                    self.createAndShowAlert('Upload your file using the form in the new window. Once you have a Geomark URL, load it using "Load an Existing Geomark".');
```
dgboss: maybe '..., add it to the map using "Load an Existing Geomark".

src/smk/tool/geomark/tool-geomark.js
dgboss: Oh, I might see what's going on. I think you can simplify things by calling self.loadGeomark directly in the 'close-prompt' handler.

dgboss (changes requested):
Let's start here and work on styling later.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I use "handlePrompt" for dynamic handling. It's unnecessary now because there's only one prompt, but if we added another, we'd want the "close prompt" event to call whichever handler function we assigned before showing the prompt.

Do you think it's overkill to keep it as-is? This could then be reused elsewhere going forward. Or should we trust that we can implement "close prompt" handling in a reusable way if/when we need to going forward?

src/smk/component/alert/component-alert.css
michaelpnelson: Screen Shot 2021-10-29 at 11 18 24 AM

src/smk/component/alert/component-alert.css
michaelpnelson: 400px looks like a more natural width to me, so I'll drop this from 500px in the next change set.

I'll add max-width: 90% as well, though I don't see its effect when trying it out locally - I'd expect it to be a bit narrower.

src/smk/component/alert/component-alert.css
michaelpnelson: Sounds good. I'll also apply the "smk-" prefix I see in other CSS.

src/smk/component/alert/component-alert.css
michaelpnelson: Done ... I've restored the "vertical-align: middle" property.

src/smk/component/alert/component-alert.css
michaelpnelson: I think the reason these alert and prompt components appear behind the toolbar is that the components belong to the panel, which will have a lower z-index than the toolbar. 

One suggestion would be to add an alert and a prompt component each to an element that's higher in the stack, for reuse by all tools/panels. 

 Now that the modal wrapper is centre-aligned vertically, there's no longer overlap between the modal wrapper and the toolbar box.

src/smk/component/prompt/component-prompt.html
michaelpnelson: Specifying margin-left on the button class is working well.

src/smk/component/alert/component-alert.html
michaelpnelson: "close-alert" is already used as a name for an emit from panel-geomark.html to tool-geomark.js. I can make this (the emit from component-alert.html to panel-geomark.html) to be more specific, but I'm curious - wouldn't the emit be scoped to the component, and then not be susceptible to collision with an emit of the same name in a different component (the prompt)?

src/smk/component/alert/component-alert.js
michaelpnelson: No; I'll remove it.

src/smk/component/prompt/component-prompt.html
michaelpnelson: Done.

src/smk/component/prompt/component-prompt.html
michaelpnelson: Done

src/smk/component/prompt/component-prompt.html
michaelpnelson: Thanks. I think v-model is cleaner and will use it. I already used it in panel-geomark.html but didn't think to use it here.

src/smk/component/alert/component-alert.css
dgboss: This behavior must differ between browsers or platforms, I don't see any issue on Windows in Chrome, FF or Edge.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I like "updateAndShowAlert" better - "showAlert" is already overloaded as the name of a variable.

src/smk/tool/geomark/tool-geomark.js

src/smk/component/alert/component-alert.css
dgboss: The 90% looks decent in my testing.

src/smk/component/alert/component-alert.css
dgboss: Looks good.

src/smk/component/alert/component-alert.html
```diff
@@ -0,0 +1,15 @@
+<div class="smk-modal-mask">
+  <div class="smk-modal-wrapper">
+    <div class="smk-modal-container">
+      <div class="smk-modal-body">
+        <slot name="alertBody"></slot>
+      </div>
+      <div class="smk-modal-footer">
+        &nbsp;
```
dgboss: This non-breaking space should be removed as well.

src/smk/tool/geomark/tool-geomark.js
dgboss: We're highly unlikely to require an additional prompt in the Geomark component. I'd recommend simplifying in this case.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -151,33 +205,13 @@ include.module( 'tool-geomark', [
                     });
                 },
                 'create-geomark-from-file': function () {
-                    alert('Upload your file using the form in the new window. Once you have a Geomark URL, load it using "Load an Existing Geomark".');
+                    self.updateAndShowAlert('Upload your file using the form in the new window. Once you have a Geomark URL, add it to the map using "Load an Existing Geomark".');
```
dgboss, line 214: With the custom alert, we've now lost the desirable behavior where the app blocks until the user clicks ok and we then open the window.

src/smk/component/alert/component-alert.css
dgboss: Yeah, probably best setting the font size to 1em or 1rem for both components, though changes to the showStatusMessage prompt are out of scope for this PR, so I'd leave that one alone for now.

src/smk/tool/geomark/tool-geomark.js
dgboss: Good idea to try and anticipate future needs! I do think it is highly unlikely we will add an additional prompt to the geomark component though. My personal preference would be for simplicity in this case.

src/smk/component/alert/component-alert.html
michaelpnelson: The remaining non-breaking space is deliberate, to give us desired vertical spacing. If it's removed, the alert looks like:

Screen Shot 2021-11-02 at 8 28 44 AM

I wasn't sure how to address it in CSS, but on further thought, setting a fixed height on the "smk-modal-footer" div seems like a reasonable fix as we probably wouldn't expect variable text (or any text) in that area. Actually, on even further thought, min-height seems safer, in case we do get text added to the footer.

src/smk/component/alert/component-alert.css
michaelpnelson: To my eye, setting a font size of 1em or 1rem appears the same as the inherited font size, so I'll plan to just remove the font-size here.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I'll remove the "handlePrompt" variable then.

src/smk/component/alert/component-alert.html
dgboss: I'm not sure how Vue scopes events and whether or not you need to manually prevent propagation. It's also helpful to have descriptive events.

src/smk/component/alert/component-alert.css
dgboss: Looks good.

src/smk/component/alert/component-alert.css
dgboss: This works fine, but there is something to be aware of for future reference. A more common, modern approach for centering content could be accomplished by adding css like the following to the .smk-modal-container class, removing the display: table-cell property and getting rid of the smk-modal-wrapper div and css altogether.

    position: absolute;
    top: 50%;
    left: 50%;
    -moz-transform: translateX(-50%) translateY(-50%);
    -webkit-transform: translateX(-50%) translateY(-50%);
    transform: translateX(-50%) translateY(-50%);

Just food for thought, no changes necessary.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I've added dynamic handling of alerts so that we can conditionally open the "create geomark from file" window when closing an alert. I looked into a callback but it seems that would involve passing the callback function through the events, which I'm not sure we could do.

dgboss: Looks good.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -102,6 +120,13 @@ include.module( 'tool-geomark', [
                 }
             } )
 
+            this.updateAndShowAlert = function(alertBodyArg) {
+                self.alertBody = alertBodyArg;
+                self.showAlert = true;
+            }
+
+            this.handleAlert = function() {};
```
dgboss: JavaScript has no concept of typings, so you don't need to assign an empty function to this property. This results in some additional overhead (though very small) that isn't necessary. You can simply set `this.handleAlert = undefined` here and on line 233.

dgboss: And maybe add a comment stating the purpose of the handleAlert property.

dgboss (approved):
I'm approving now, but please see my last two comments regarding a couple final small changes.

src/smk/tool/geomark/tool-geomark.js

PR #137, opened 2021-11-04
Status message inherits font size

michaelpnelson:
This removes the smaller font size set for the status message component, to better match the new modal components. It also removes commented-out styles.


dgboss (approved):
Looks good!

PR #138, opened 2021-11-05
Proof of concept: different path styles for Markup, Geomark tools

michaelpnelson:
This demonstrates the ability to have unique path styles for a tool.

The Markup tool is treated as a default draw tool. It is updated to set default styles for the temp line and hint line. The Geomark tool is updated to capture these styles when entered and set different styles; on exit, it restores the default styles.

Darren, please let me know what you think of this approach. 

I'm not sure it's worth the effort as it adds complexity that we'd have to maintain going forward, and could potentially complicate other future changes.

### DIFF
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -109,14 +109,24 @@ include.module( 'tool-geomark', [
                 }
             } 
 
+            this.defaultTemplineStyle = undefined;
+            this.defaultHintlineStyle = undefined;
             this.changedActive( function () {
                 if ( self.active ) {
+                    if (!this.defaultTemplineStyle) {
+                        this.defaultTemplineStyle = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().templineStyle));
+                    }
+                    if (!this.defaultHintlineStyle) {
+                        this.defaultHintlineStyle = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().hintlineStyle));
+                    }
+                    smk.$viewer.map.pm.setGlobalOptions({ templineStyle: { color: 'black' }, hintlineStyle: { color: 'green' } });
                     smk.$viewer.map.pm.enableDraw('Polygon', {
                         continueDrawing: true
                     });
                 }
                 else {
                     smk.$viewer.map.pm.disableDraw();
+                    smk.$viewer.map.pm.setGlobalOptions({ templineStyle: this.defaultTemplineStyle, hintlineStyle: this.defaultHintlineStyle });
                 }
             } )
 
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -110,23 +110,16 @@ include.module( 'tool-geomark', [
             } 
 
             this.setCurrentDrawingLayer = function(e) {
-                currentDrawingLayer.addLayer(e.layer);
+                var eventLayer = e.layer;
+                eventLayer.pm.setOptions( {
+                    'allowEditing': false,
+                    'allowRemoval': false
+                } );
+                currentDrawingLayer.addLayer(eventLayer);
             }
 
-            this.defaultTemplineStyle = undefined;
-            this.defaultHintlineStyle = undefined;
-            this.defaultPathOptions = undefined;
             this.changedActive( function () {
                 if ( self.active ) {
-                    if (!this.defaultTemplineStyle) {
-                        this.defaultTemplineStyle = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().templineStyle));
-                    }
-                    if (!this.defaultHintlineStyle) {
-                        this.defaultHintlineStyle = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().hintlineStyle));
-                    }
-                    if (!this.defaultPathOptions) {
-                        this.defaultPathOptions = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().pathOptions));
-                    }
                     smk.$viewer.map.pm.setGlobalOptions({ 
                         templineStyle: { 
                             color: '#003366' 
@@ -147,12 +140,8 @@ include.module( 'tool-geomark', [
                 }
                 else {
                     smk.$viewer.map.pm.disableDraw();
-                    smk.$viewer.map.on('pm:create', undefined); // FIXME polygons drawn by Markup still trigger self.setCurrentDrawingLayer 
-                    smk.$viewer.map.pm.setGlobalOptions({ 
-                        templineStyle: this.defaultTemplineStyle, 
-                        hintlineStyle: this.defaultHintlineStyle,
-                        pathOptions: this.defaultPathOptions
-                    });
+                    smk.$viewer.map.off('pm:create', self.setCurrentDrawingLayer);
+                    self.setDefaultDrawStyle();
                 }
             } )
 
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -122,15 +122,15 @@ include.module( 'tool-geomark', [
                 if ( self.active ) {
                     smk.$viewer.map.pm.setGlobalOptions({ 
                         templineStyle: { 
-                            color: '#003366' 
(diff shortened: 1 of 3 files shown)


src/smk/tool/geomark/tool-geomark.js
```diff
@@ -109,14 +109,24 @@ include.module( 'tool-geomark', [
                 }
             } 
 
+            this.defaultTemplineStyle = undefined;
+            this.defaultHintlineStyle = undefined;
             this.changedActive( function () {
                 if ( self.active ) {
+                    if (!this.defaultTemplineStyle) {
```
michaelpnelson: The true "default" path styles (the blue colour and so on) doesn't seem to be exposed in global options; they only seem to contain custom values that have been set.

This is also dependent on the styles being set in the Markup tool first, though this could be rewritten to be more defensive if styles have not been set. 

This could also be optimized to scale better (should we want to set more styles) by pulling it out to a function.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -109,14 +109,24 @@ include.module( 'tool-geomark', [
...
+                        this.defaultTemplineStyle = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().templineStyle));
```
michaelpnelson: I'm using the JSON methods to capture the value of the style object - otherwise we only have a reference and we can't easily reset the style to the default. Let me know if you have better suggestions for capturing the value.

src/smk/tool/geomark/tool-geomark.js
dgboss: Yeah, we definitely need to be more defensive. What do you think about reading values from config if available, and falling back to reasonable hard coded defaults?

src/smk/tool/geomark/tool-geomark.js
dgboss: This is great!

dgboss:
I think your approach is great and definitely worth the effort. It looks great! The additional complexity is fairly small and quite manageable I think. A few more changes will likely be necessary, including suitable defaults as mentioned above. I think you'll also need to enable/disable the pm:create event listener so we don't treat regular markup as geomark drawings. We probably need to disable edit capabilities on geomark layers as well so a user can't use the markup tools to edit geomark shapes.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: When you suggest reading values from "config", do you mean smk-config.json? I'd expect to add the config to the Markup tool then. 

Alternately, the default values could go into a function that is visible to all tools. I'm a bit concerned about having CSS rules defined in JS code; maybe we could find a way to read values from a CSS file.

michaelpnelson: Correcting myself: Path options look like CSS but aren't, so I'm less concerned about defining them in JS code. Leaflet path options are defined in https://leafletjs.com/reference.html#path.

dgboss: Oh, sorry for the lack of clarity. I'm thinking faster than I can type! I'm thinking hard coded default color for the regular markup tools, and reading the colors for the Geomark module from config. Perhaps a little overkill for the feature though. It could always be done as an enhancement at some point in the future.

Would it be possible to allow the geomark and markup modules to be 'greedy'? ie. Set the color of their choosing in the this.changedActive function and not worrying about restoring previous values?

michaelpnelson: Thanks for clarifying. 

About reading Geomark styles from config ... I'm agreeing that may be overkill. We'd have to do some design and discussion on which style options we'd want to be able to set, and how to set them.

About being greedy ... I don't think the Markup tool could be greedy because it's not a panel and can't know whether it's active to override any previously set style options. That's why my implementation sets the default in Markup, and then Geomark (and any other future panel-based tools that draw shapes) would need to set and unset style options on active and inactive events.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -119,14 +124,35 @@ include.module( 'tool-geomark', [
...
+                    });
+                    smk.$viewer.map.on('pm:create', self.setCurrentDrawingLayer);
                     smk.$viewer.map.pm.enableDraw('Polygon', {
                         continueDrawing: true
                     });
                 }
                 else {
                     smk.$viewer.map.pm.disableDraw();
-                    smk.$viewer.map.pm.setGlobalOptions({ templineStyle: this.defaultTemplineStyle, hintlineStyle: this.defaultHintlineStyle });
+                    smk.$viewer.map.on('pm:create', undefined); // FIXME polygons drawn by Markup still trigger self.setCurrentDrawingLayer
```
michaelpnelson: I'm looking into this, but if you see an obvious problem please let me know.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -119,14 +124,35 @@ include.module( 'tool-geomark', [
                     if (!this.defaultHintlineStyle) {
                         this.defaultHintlineStyle = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().hintlineStyle));
                     }
-                    smk.$viewer.map.pm.setGlobalOptions({ templineStyle: { color: 'black' }, hintlineStyle: { color: 'green' } });
+                    if (!this.defaultPathOptions) {
+                        this.defaultPathOptions = JSON.parse(JSON.stringify(smk.$viewer.map.pm.getGlobalOptions().pathOptions));
+                    }
+                    smk.$viewer.map.pm.setGlobalOptions({ 
+                        templineStyle: { 
+                            color: '#003366'
```
michaelpnelson: This is a different colour to differentiate Geomark-drawn shapes from Markup-drawn shapes. I'm going with a pretty arbitrary but not too inappropriate choice - it's the background deep-blue colour used in other BC pages. It looks more like near-black to me when drawn over a map. Let me know if you have other suggestions.

src/smk/tool/geomark/tool-geomark.js
dgboss: AH, good point! I forgot about that. Also, there is almost no code in the markup tool to modify.

src/smk/tool/geomark/tool-geomark.js
dgboss: I don't think this is quite the fix/behaviour we're looking for, but you have the right idea. In this else block, you'll want to unsubscribe from listening for the 'pm:create' event.

I see you are subscribing to the event with: 
`smk.$viewer.map.on('pm:create', self.setCurrentDrawingLayer);`

You can try unsubscribing with:
`smk.$viewer.map.off('pm:create', self.setCurrentDrawingLayer);`

There is a chance this won't work. You might actually need to define setCurrentDrawingLayer like this instead:
`var setCurrentDrawingLayer = function(){ ... }`

Your subscribe and unsubscribe would then look like:
`smk.$viewer.map.on('pm:create', setCurrentDrawingLayer);`
`smk.$viewer.map.off('pm:create', setCurrentDrawingLayer);`

michaelpnelson: Thanks! Your first suggestion is working. I wasn't aware of the "off" method - that's great to know.

src/smk/tool/tool-base.js
```diff
@@ -126,6 +126,22 @@ include.module( 'tool.tool-base-js', [ 'tool.tool-js' ], function ( inc ) {
                 return smk.getStatusMessage().show( message, status, delay, this.busy )
             }
 
+            this.setDefaultDrawStyle = function() {
+                smk.$viewer.map.pm.setGlobalOptions({ 
+                    templineStyle: { 
+                        color: 'red',
```
dgboss, line 132: Nice! I like that you put this in the base class/module!

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: This is a test reply

src/smk/tool/geomark/tool-geomark.js
dgboss: I think we'll want to choose something with a little more contrast. Perhaps we can discuss in the meeting later this morning.

dgboss (approved):

michaelpnelson:
The Geomark style colour has been changed to ee0077 (rounded from the e00275 value we experimented with in our review meeting). 

Since this addresses the last review comment, I'll go ahead and merge this.

PR #139, opened 2021-11-08
Create From File link is configurable

michaelpnelson:
This adds support for a "enableCreateFromFile" configuration property. When set true, the "Create geomark from a file" link is shown; otherwise, it is hidden.

Based on customer feedback, the existing link is not of high value, but we want app authors to be able to include it if they feel it is helpful to their users.

Example configuration for the Geomark tool:

        {
            "title": "Geomark",
            "geomarkService": {
                "url": "https://apps.gov.bc.ca/pub/geomark"
            },
            "enableCreateFromFile": true,
            "type": "geomark",
            "enabled": true,
        },

### DIFF
--- a/docs/configuration/tools/geomark.md
+++ b/docs/configuration/tools/geomark.md
@@ -45,4 +45,4 @@ The URL of the geomark web service used.
 ## Enable Create From File Property
 `"enableCreateFromFile"`: `Boolean`
 
-When set to true, a link will be added to the panel that opens a new browser tab to the page of the geomark service where a file can be uploaded to create a new geomark. The geomark's link can then be used to load a geomark.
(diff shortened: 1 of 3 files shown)


dgboss (approved):
The code looks good. Did you want to update the documentation as part of this PR as well?

michaelpnelson:
I'll add another change updating the documentation. I did it for smk-cli (locally) but forgot to do it here :) .

michaelpnelson:
Darren, you've approved this, but I'd appreciate it if you could take a quick look at the update to the docs and let me know whether you have changes to suggest.

michaelpnelson:
The text of links will be changing based on feedback, so I'd prefer not to quote the link text and instead just mention "load" functionality.

How does this sound?

"When true, the panel will contain a link to the Geomark service where a file can be uploaded to create a new geomark. The geomark can then be added to the map by loading it."

dgboss (approved):
Sounds good

PR #140, opened 2021-11-16
Markup toolbar removed when Geomark panel is active

michaelpnelson:
This removes the draw controls used by the Markup toolbar when the Geomark panel becomes active. When the Geomark panel is inactive, the Markup toolbar is restored.

Code for toolbar options has been moved from Markup code to tool base code so other tools like Geomark can call it.

### DIFF
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -120,9 +120,14 @@ include.module( 'tool-geomark', [
                 currentDrawingLayer.addLayer(eventLayer);
             }
 
+            this.toggleMarkupToolbarControls = function() {
+                if (smk.$tool.MarkupTool) {
+                    smk.$viewer.map.pm.toggleControls();
+                }
+            }
+
             this.changedActive( function () {
                 if ( self.active ) {
-                    smk.$viewer.map.pm.removeControls(); // remove Markup toolbar
                     smk.$viewer.map.pm.setGlobalOptions({ 
                         templineStyle: { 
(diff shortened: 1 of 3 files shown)


src/smk/tool/geomark/tool-geomark.js
```diff
@@ -122,6 +122,7 @@ include.module( 'tool-geomark', [
 
             this.changedActive( function () {
                 if ( self.active ) {
+                    smk.$viewer.map.pm.removeControls(); // remove Markup toolbar
```
dgboss: Rather than removing and re-adding the toolbar each time (which requires you to ensure config is in sync with other code), maybe take a look at the 'toggleControls' command.

dgboss (changes requested):
I though of one other option you could explore, depending on how you feel about the user experience with respect to hiding/showing controls. It is also possible to disable the buttons on the toolbar using the 'setButtonDisabled' command. This option might be more visually appealing. What do you think?

michaelpnelson:
The approach of disabling and re-enabling toolbar buttons works fine but for one thing. When the Geomark panel is active, we enable draw mode with the Polygon tool, and as a result the Polygon tool is selected. This causes the "helper" buttons (Finish, Remove Last Vertex, Cancel) to slide out from under the Polygon tool. With the Polygon tool disabled, we can't make these helper buttons go away via Cancel.

I haven't found a way to hide these helper buttons via Toolbar or Draw Mode API. I'll push a change that toggles the toolbar controls instead.

dgboss (approved):
Looks good

PR #141, opened 2021-11-16
Close-alert error fixed

michaelpnelson:
This addresses the error (seen in console logs) given when attempting to call the "handleAlert" function when it is undefined. This adds a check that the function is defined before calling it.

To reproduce the error, draw a geomark shape, save the drawing, and click 'OK' on the alert that appears afterwards.


dgboss (approved):
Good catch

PR #142, opened 2021-11-18
Command links are replaced with buttons

michaelpnelson:


src/smk/tool/geomark/panel-geomark.html
```diff
@@ -7,14 +7,14 @@
...
-    </ul>
+    <div>
+        <button v-on:click="$$emit( 'create-geomark-from-drawing' )" :disabled="shapeIsDrawn === false">Save</button>
+        <button v-on:click="$$emit( 'clear-drawing' )" :disabled="shapeIsDrawn === false">Clear</button>
+        <button v-on:click="$$emit( 'load-geomark' )">Load URL</button>
+        <button v-if="enableCreateFromFile === true" v-on:click="$$emit( 'create-geomark-from-file' )">Load File</button>
+    </div>
 
-    <div v-if="geomarks.length != 0" id="geomarks-container">
+    <div v-if="geomarks.length !== 0" id="geomarks-container">
```
michaelpnelson, line 17: This isn't related but I thought I'd update this while I noticed that I've been using === for equality but != for inequality. I fixed it elsewhere as well.

dgboss:
The changes look good, though I did come up with one use case that is a little odd. Try the following:
1. Enable the Geomark tool/panel
2. Draw a polygon on the map, but don't save it
3. Close the geomark panel
4. Re-open the geomark panel. I would expect the buttons to still be active since I have an unsaved geomark on my map, but they are inactive.

dgboss (changes requested):
The changes look good, though I did come up with one use case that is a little odd. Try the following:

1. Enable the Geomark tool/panel
2. Draw a polygon on the map, but don't save it
3. Close the geomark panel
4. Re-open the geomark panel. I would expect the buttons to still be active since I have an unsaved geomark on my map, but they are inactive.

michaelpnelson:
> The changes look good, though I did come up with one use case that is a little odd. Try the following:
> 
> 1. Enable the Geomark tool/panel
> 2. Draw a polygon on the map, but don't save it
> 3. Close the geomark panel
> 4. Re-open the geomark panel. I would expect the buttons to still be active since I have an unsaved geomark on my map, but they are inactive.

Good catch!

The "shapeIsDrawn" variable is set to false on initialization, but it's also set to false when the panel becomes active. I'll remove this line and push another commit.

dgboss (approved):
Looks good!

PR #143, opened 2021-11-22
Custom style applied to loaded geomark

michaelpnelson:
This addresses an overlooked issue where a loaded geomark's shape uses the default style instead of the custom style for geomarks.

### DIFF
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -51,6 +51,8 @@ include.module( 'tool-geomark', [
         function ( smk ) {
             var self = this;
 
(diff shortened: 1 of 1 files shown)


src/smk/tool/geomark/tool-geomark.js
```diff
@@ -51,6 +51,8 @@ include.module( 'tool-geomark', [
         function ( smk ) {
             var self = this;
 
+            var CUSTOM_COLOUR = '#ee0077';
```
michaelpnelson, line 54: I wanted to make this a const, but this gives a warning as we didn't set ES6 support. (Same for "let".)

dgboss (approved):
Nice!

PR #144, opened 2021-11-24
Geomark is disabled for mobile devices

michaelpnelson:


michaelpnelson:
Some notes from my testing ...

When an SMK app is created with "mobile" specified, the geomark panel doesn't appear, even when enabled in smk-config.json; I believe this is because geomark is not included in the built app (I don't see it among the node module files).

When an SMK app is created with "default" specified, the geomark panel appears for a mobile device, but there's no functionality when it's selected - we don't enter draw mode and buttons don't work. 

We disable other tools this way (coordinate, zoom, markup, minimap) but this may be the first time we've disabled a panel this way. I didn't see a way to exclude the panel from the list of tools for mobile devices; we test on the smk.$device variable, but it looks like we don't have a handle on the smk object until after tools are initialized.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -52,6 +52,10 @@ include.module( 'tool-geomark', [
             this.promptReply = '';
         },
         function ( smk ) {
+            if ( smk.$device == 'mobile' ) {
```
dgboss: We should use === here.

dgboss (changes requested):

dgboss (changes requested):
Also consider updating documentation to indicate that the geomoark tool isn't supported when device === mobile.

michaelpnelson:
> Also consider updating documentation to indicate that the geomoark tool isn't supported when device === mobile.

Done.

src/smk/tool/geomark/tool-geomark.js

dgboss (approved):
Looks good.

PR #145, opened 2021-11-24
Loaded geomarks cannot be edited or removed

michaelpnelson:
This addresses an oversight where new geomarks were made immutable but loaded geomarks were not.


dgboss (approved):
Nice catch and good use of the word immutable!

PR #147, opened 2021-11-26
Geomark feature

michaelpnelson:


dgboss (changes requested):
Just one quick thought... what do you think about making the "Loaded Geomarks' title an H3 instead of an H2? The H2 version is very 'loud' in my opinion.

michaelpnelson:
> Just one quick thought... what do you think about making the "Loaded Geomarks' title an H3 instead of an H2? The H2 version is very 'loud' in my opinion.

H3 looks fine to me. We have some H2s and H3s in other panels; making this an H3 doesn't seem out of place.

I'll push another commit to this geomark_feature branch with the change.

dgboss (approved):
DONE!

PR #149, opened 2021-11-29
Updating package-lock for 1.0.7

michaelpnelson:


michaelpnelson:
I forgot to do an npm install to update the package lock ...

On further inspection, this change adds many more lines than it should. I'll abandon this and push another pull request with just the version change.

PR #151, opened 2021-11-30
Improved cleanup on Geomark exit

michaelpnelson:
This updates the logic on whether we enable the Save button by checking that the current layer group has layers.

It also turns off some event handling on panel exit that was previously left on and triggerable by other use of the events (i.e. the Markup tool).

### DIFF
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -58,14 +58,12 @@ include.module( 'tool-geomark', [
             this.isMobile = false;
         },
         function ( smk ) {
-            if ( smk.$device === 'mobile' ) {
-                this.isMobile = true;
-                return;
-            }
-
             var self = this;
 
-            var CUSTOM_COLOUR = '#ee0077';
+            if ( smk.$device === 'mobile' ) {
+                self.isMobile = true;
+                return;
+            }
 
             // Check for "geomarkService" configuration. Example:
             // "geomarkService": {
@@ -76,11 +74,20 @@ include.module( 'tool-geomark', [
                 return;
             }
 
-            this.createCurrentDrawingLayer = function() {
-                var drawingLayer = new L.FeatureGroup();
-                smk.$viewer.map.addLayer(drawingLayer);
-                return drawingLayer;
+            var CUSTOM_COLOUR = '#ee0077';
+
+            // Used to specify action(s) to be executed when an alert is confirmed
+            this.handleAlert = undefined;
+
+            this.createCurrentLayerGroup = function() {
+                var layerGroup = new L.FeatureGroup();
+                smk.$viewer.map.addLayer(layerGroup);
+                return layerGroup;
             }
+
+            var currentLayerGroup = self.createCurrentLayerGroup();
+
+            var currentLayer;
 
             this.buildWktCoords = function(layerGroup) {
                 var isMultiPolygon = layerGroup.getLayers().length > 1;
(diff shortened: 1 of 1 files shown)


src/smk/tool/geomark/tool-geomark.js
```diff
@@ -140,6 +140,18 @@ include.module( 'tool-geomark', [
...
             }
 
+            this.drawStart = function(e1) {
+                e1.workingLayer.on('pm:vertexadded', function(e2) {
+                    self.canClear = true;
+                });
+            }
+
+            this.drawEnd = function(e) {
+                if (currentDrawingLayer.getLayers().length > 0) {
```
dgboss: Is getLayers() guaranteed to return something? Or is it possible for it to return undefined? If it can return undefined this will throw an error.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -140,6 +140,18 @@ include.module( 'tool-geomark', [
                 currentDrawingLayer.addLayer(eventLayer);
             }
 
+            this.drawStart = function(e1) {
+                e1.workingLayer.on('pm:vertexadded', function(e2) {
```
dgboss: Since you're doing some cleanup, we should unsubscribe from the pm:vertexadded event as well.

dgboss (changes requested):
Good idea to do some of this cleanup!

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: We already unsubscribe this function from the 'pm:drawstart' event, so my thinking (and reading https://leafletjs.com/reference.html#map-off) is that this would be sufficient, because now map is no longer listening for this event with this function. Also, I'm not sure how we'd unsubscribe 'pm:vertexadded' events and function because we don't have a reference on the layers at this point. Let me know if I'm missing something.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: LayerGroup.getLayers() (https://leafletjs.com/reference.html#layergroup-getlayers) returns an array. I also tested with an "if (!currentDrawingLayer.getLayers())" block and confirmed it doesn't get entered.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: I also forgot to mention this earlier ... I also tested with a breakpoint in the function that we subscribe 'pm:vertexadded' on, and confirmed that when I leave the Geomark panel and draw shapes using Markup, the breakpoint is not called.

dgboss: Any time you subscribe to an event you really should be unsubscribing as well, or else you can end up with memory leaks. Probably not a huge issue for SMK and the geomark tool, but still a valid pattern to follow. Here's how you accomplish this:
1. create a new variable (eg. var activeDrawingLayer) to hold a reference to the 'working layer'
2. Extract the pm:vertexadded event handler (ie. the inline anonymous function(e2)) to a named function (eg. this.setCanClear())
3. Set this new function as the event handler for the pm:vertexadded event. (eg e1.workingLayer.on('pm:vertexadded', self.setCanClear))
4. In this.drawStart assign e1.workingLayer to your new variable so you have a reference to the layer that the drawing tool is using
5. In this.drawEnd unsubscribe (eg. activeDrawingLayer.off('pm:vertexadded',  self.setCanClear)

src/smk/tool/geomark/tool-geomark.js
dgboss: Great, thanks for checking!

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Thanks for this suggestion! I've implemented it and it tests well.

Because "activeDrawingLayer" is similar to the existing variable "currentDrawingLayer", I'm renaming the latter as "currentLayerGroup".

dgboss (approved):

src/smk/tool/geomark/tool-geomark.js
dgboss: Looks great!

PR #155, opened 2021-12-08
Add Cypress tests

michaelpnelson:
This adds the Cypress testing library and a few basic tests. This also tests some code in util.js; going forward, testable code can be added to util.js.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,3 @@
+{
+    "video": false
+}
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
--- a/cypress/integration/basic_spec.js
+++ b/cypress/integration/basic_spec.js
@@ -1,0 +1,17 @@
+describe('SMK Basic Tests', () => {
+    it('Has a non-empty SMK object', () => {
+        cy.window().its('SMK').should('not.be.null');
+    })
+
+    it('Has a non-empty SMK.MAP object', () => {
+        cy.window().its('SMK.MAP').should('not.be.null');
+    })
+
+    it('Has the expected Map name', () => {
+        cy.window().its('SMK.MAP.1.name').should('eq', 'SMK Default Map');
+    })
+
+    it('Has a Location tool with the expected ID', () => {
+        cy.window().its('SMK.MAP.1.$tool.LocationTool.rootId').should('eq', 'LocationTool');
+    })
+})
--- a/cypress/integration/util_spec.js
+++ b/cypress/integration/util_spec.js
@@ -1,0 +1,38 @@
+describe('SMK Util Tests', () => {
+    it('Has a non-empty SMK.UTIL object', () => {
+        cy.window().its('SMK.UTIL').should('not.be.null');
+    })
+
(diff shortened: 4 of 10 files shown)


michaelpnelson:
This gives us the ability to add automated tests going forward. 

We could eventually create a Github Action to trigger tests, but I think this is premature and ad-hoc testing should be fine for now.

cypress.json
```diff
@@ -0,0 +1,3 @@
+{
+    "video": false
```
michaelpnelson, line 2: When tests fail, failure artifacts are added to cypress/screenshots and cypress/videos directories. This line disables capturing videos.

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
michaelpnelson, line 1: This example file is regenerated if the fixtures directory is empty. We could alternately add it to .gitignore.

cypress/integration/util_spec.js
```diff
@@ -0,0 +1,38 @@
+describe('SMK Util Tests', () => {
+    it('Has a non-empty SMK.UTIL object', () => {
+        cy.window().its('SMK.UTIL').should('not.be.null');
+    })
+
+    it('Should get meters per millimeter', () => {
+        let util;
+        cy.window()
```
michaelpnelson, line 8: I tried a few ways to pull out this code since it's repeated in the test below, but I don't think I understand Promises well enough to get it working. I also found that cy.window() cannot be called outside an "it" block.

cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
```
michaelpnelson, line 1: This is also generated. I saw it was also included in leaflet-geoman.

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
michaelpnelson, line 1: This is also generated ...

cypress/support/index.js
```diff
@@ -0,0 +1,25 @@
...
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
+
+beforeEach(() => {
```
michaelpnelson, line 22: This is also generated but for this line and below. Leaflet-geoman kept the boilerplate above so I did as well.

cypress/support/index.js
```diff
@@ -0,0 +1,25 @@
...
+    cy.visit('./index.html');
```
michaelpnelson, line 23: It's a bit of overhead to add the index.html file. I considered using one of the apps in /debug, but those take considerably longer to load.

index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
```
michaelpnelson, line 1: This is copied from the SMK Github page :)

cypress/integration/basic_spec.js
```diff
@@ -0,0 +1,17 @@
+describe('SMK Basic Tests', () => {
```
dgboss, line 1: Maybe rename file to indicate what is being tested? smk_spec or smk_map_spec?

dgboss (changes requested):

michaelpnelson:
Closing this. We want to focus on testing individual modules over testing against SMK as a whole.

PR #156, opened 2021-12-08
Override Identify click handling

michaelpnelson:
This updates the Geomark tool to override the default click handling of the Identify tool, but only when the Geomark panel is active.

### DIFF
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -202,6 +202,14 @@ include.module( 'tool-geomark', [
                 }
             } )
 
+            // Override default handling in Identify tool
+            smk.$viewer.handlePick( 1, function ( location ) {
+                if ( !self.active ) {
+                    return;
+                }
+                return true;
+            } )
+
             this.updateAndShowAlert = function(alertBodyArg) {
                 self.alertBody = alertBodyArg;
                 self.showAlert = true;
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -207,7 +207,7 @@ include.module( 'tool-geomark', [
                 if ( !self.active ) {
                     return;
                 }
-                return true;
+                return Promise.resolve(true);
             } )
 
             this.updateAndShowAlert = function(alertBodyArg) {
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -203,7 +203,7 @@ include.module( 'tool-geomark', [
             } )
 
             // Override default handling in Identify tool
-            smk.$viewer.handlePick( 1, function ( location ) {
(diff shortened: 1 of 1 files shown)


michaelpnelson:
An update: I found I could detect when Markup is drawing by listening to pm:drawstart and pm:drawend events. Unfortunately we don't know when to stop listening to those events because we don't exit a panel, so it keeps handling the same events for Geomark.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -202,6 +202,14 @@ include.module( 'tool-geomark', [
                 }
             } )
 
+            // Override default handling in Identify tool
+            smk.$viewer.handlePick( 1, function ( location ) {
+                if ( !self.active ) {
+                    return;
+                }
+                return true;
```
dgboss: It looks like the Identify tool actually returns a promise that resolves with a value of true. I wonder if we should follow a similar pattern?

michaelpnelson: The identity tool chains a promise onto a function, but here there isn't a function to chain onto, so I'm not sure a promise makes sense. Or am I missing something? Maybe I should return a function that returns true instead of returning true? (That seems to work as well.)

dgboss: Try replacing 'return true;' with 'return Promise.resolve(true);'. This should keep the desired behavior while also honoring the existing pattern/method signature.

michaelpnelson: Thanks for this suggestion. It works well. I'll push it soon.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -202,6 +202,14 @@ include.module( 'tool-geomark', [
                 }
             } )
 
+            // Override default handling in Identify tool
+            smk.$viewer.handlePick( 1, function ( location ) {
```
dgboss: What does this 'magic' number mean? Why is it set to 1?

michaelpnelson: The number is the priority. Tools calling handlePick() use the values 1, 2, or 3, except for Identify, which calls it twice and uses 0 for the "fallback handler". I chose 1 so it's not too high as to conflict with other tools, but greater than 0, which results in Identify being handled first.

dgboss: The Geomark tool collides with the Location tool when both are enabled. The location tool has priority 1. I'd suggest trying a priority of 4 for the Geomark tool.

michaelpnelson: Thanks for finding this ... I should have also tested with all the other tools that register click handling.

A priority of 2 is enough to resolve the collision with Location, but on further thought, I'm fine with 4 - we want to help ensure there are no conflicts with other tools going forward.

dgboss (approved):

Issue #153, opened 2021-12-01
Loaded geomarks are all checked when returning to panel


michaelpnelson:
## How to reproduce

- Create or load multiple geomarks in the Geomark tool
- Uncheck a geomark
- Navigate out of the Geomark tool
- Return to the Geomark tool
- The previously unchecked geomark is checked

michaelpnelson:
This was addressed in pull request https://github.com/bcgov/smk/pull/157.

PR #157, opened 2021-12-15
Geomark checkboxes reflect visibility

michaelpnelson:
When leaving and re-entering the Geomark panel, loaded geomarks are checked by default. This adds visibility state to members of the geomarks property, which is used to determine whether each geomark's checkbox is checked.

This addresses issue https://github.com/bcgov/smk/issues/153.


dgboss (approved):
Looks good!

PR #165, opened 2022-01-06
Add debug app for Bespoke tool

michaelpnelson:
Steps I used to run debug apps locally:
cd to debug dir
create symlink "dist" to ../dist directory:
ln -s ../dist dist
run http-server (launch npm "http-server" module)
open one of the server URLs (http://127.0.0.1:8080) in a browser
If you don't see changes after refreshing the browser, choose developer tools and ensure "Disable cache" is checked in Network tab


debug/app/bespoke/smk-config.json
```diff
@@ -0,0 +1,141 @@
...
+            "enabled": false
+        },
+        {
+            "order": 1,
+            "position": "toolbar",
+            "icon": "now_widgets",
+            "title": "Example Bespoke Tool",
+            "useComponent": true,
+            "component": {
+                "template": "<div>\n  <h3>Bespoke Tool With HTML</h3>\n  <div>\n    <p>\n      This is an example of the Bespoke tool with simple HTML\n    </p>\n  </div>\n</div>"
```
dgboss: This isn't necessary and could lead to confusion. We could actually have two Bespoke tool examples, one that demonstrates how you can add a custom template like this example and the other that uses a custom Vue component.

debug/app/bespoke/components/customComponent.js
```diff
@@ -0,0 +1,27 @@
+SMK.HANDLER.set('BespokeTool--101', 'activated', (smk, tool) => {
```
dgboss: Maybe include a comment indicating what is happening and/or how this works?
You could mention that a Bespoke tool was added to the app/smk-config and that the 'BespokeTool--101' is a concatenation of 'BespokeTool' and the instance name of the configured tool (ie. '101'). 'Activated' indicates that the handler will run when the tool is activated (versus the other options of initialized, deactivated and triggered. The third argument in the SMK.Handler.set() function is the anonymous function that executes when the BespokeTool--101 is activated.

debug/app/bespoke/components/customComponent.js
```diff
@@ -0,0 +1,27 @@
+SMK.HANDLER.set('BespokeTool--101', 'activated', (smk, tool) => {
+    tool.component = Vue.component('Zoom', {
+        data: function() {
+            return {
+                buttonText: "Fly to Victoria"
+            }
+        },
+        methods: {
+            handleZoom() {
+                const leafletMap = smk.$viewer.map;
```
dgboss: Maybe a comment stating that the smk object is a global that gives you access to everything within the smk app?

debug/app/bespoke/components/customComponent.js
```diff
@@ -0,0 +1,27 @@
...
+                const newCenter = new L.LatLng(48.45,-123.35);
+                leafletMap.flyTo(newCenter, 13);
+            }
+        },
+        template: `
+            <div>
+                <p>The Bespoke tool allows you to create your own custom tool for SMK apps. You can add HTML, CSS, and Javascript to provide unique functionality.</p>
+
+                <p>This example includes a custom <a href="https://vuejs.org/">Vue.js</a> component that uses <a href="https://leafletjs.com/">Leaflet</a> API to update the map view. Click the button below to see this in action!</p>
```
dgboss: ...component that uses the Leaflet API...

debug/app/bespoke/smk-config.json
```diff
@@ -0,0 +1,141 @@
...
+        "baseMap": "Streets"
+    },
+    "layers": [],
+    "tools": [
+        {
+            "type": "toolbar",
+            "enabled": true
+        },
+        {
+            "type": "list-menu",
```
dgboss: I think you can get rid of the list menu button since no tools are present in the list.

debug/app/index.html
```diff
@@ -4,5 +4,6 @@ <h1><a href="..">SMK Debug</a> / Apps</h1>
     <li><a href="wf">Wildfirex</a></li>
     <li><a href="translink">Translink</a></li>
     <li><a href="francophone">Francophone</a></li>
+    <li><a href="bespoke">Bespoke Tool</a></li>
```
dgboss: I'm not sure that this 'belongs' in the demo app. The other options here are all custom applications, not just a demonstration of a tool. I'm not sure that the ToolConfig section is proper either. What do you think about either putting it at the top level (on the root SMK Debug page)? Or even having a 'Custom Code' link at the top level and then the Bespoke tool demo nestled in there?

dgboss (changes requested):

debug/app/index.html
michaelpnelson: I chatted with Nicole about this and she agrees about moving this out of the "app" area. One suggestion we talked about is to move it into the Features/Tool Config area, where code is shown next to an app using the tool.

debug/app/bespoke/smk-config.json
michaelpnelson: In the next patch set I went with an HTML/JS-only solution (no Vue) for simplicity.

debug/feature/Tool-Config/bespoke.html
```diff
@@ -0,0 +1,55 @@
...
+                        "type": "bespoke",
+                        "instance": "101",
+                        "enabled": true
+                    }
+                ]
+            }, '?' )
+        } )
+
+        function fly() {
+            const newCenter = new L.LatLng(49.89,-119.51);
```
michaelpnelson: I'll add some comments to this code in the next patch set.

debug/app/bespoke/smk-config.json
dgboss: I think we should also be demonstrating how a user can create a custom Vue component and wire that into the Bespoke tool. I think it would be good to keep this customization that you added as this option could definitely be useful for some people, but I feel strongly that we should demonstrate the more 'powerful' option as well. Thoughts?

debug/feature/Tool-Config/bespoke.html
```diff
@@ -0,0 +1,65 @@
...
+                        "useComponent": true,
+                        "component": {
+                            "template": 
+                            `
+                            <!-- All of a template's content must be inside a 
+                            pair of <div> tags. -->
+                            <div>
+                                <p>
+                                    The Bespoke tool allows you to create your own 
+                                    custom panels in SMK apps. You can add HTML, CSS,
```
dgboss: If we're mentioning CSS we should probably demonstrate how this is done.

debug/feature/Tool-Config/bespoke.html
```diff
@@ -0,0 +1,65 @@
...
+                        "order": 1,
+                        "position": "toolbar",
+                        "icon": "now_widgets",
+                        "title": "Bespoke Tool",
+                        "useComponent": true,
+                        "component": {
+                            "template": 
+                            `
+                            <!-- All of a template's content must be inside a 
+                            pair of <div> tags. -->
```
dgboss: Not quite true, they don't have to be div tags, there just needs to be a single set of enclosing parent tags (for example, you might be using a component here).

dgboss:
One thought about 'placement' of the Bespoke example...it is a little buried. If we go with a couple of examples of Bespoke tool customization, what do you think about following the pattern used for Identify, Layers and Search?
image

debug/app/bespoke/smk-config.json
michaelpnelson: I'm fine adding a Bespoke example having a Vue component, but it would need to be outside the Features area. Features expect all examples to be single-page web apps and to only display the source of the html page (via the addExample() method in /debug/feature/lib/examples.js). 

One solution would be to restore the app with a Vue-based bespoke tool to the Apps area, keep the single-page bespoke example where it is in the Features/Tool Config area, and add descriptive links to each other, so if you're looking at one, you are made aware of and can quickly navigate to the other. How does that sound?

dgboss: My vote would be to have the Bespoke examples side-side-by side in a 'Customization' category off the main debug landing page...
image

Upon further reflection, I'm not sure we should include the simpler example. While it does work, it is demonstrating a develop anti-pattern that I don't feel we should be promoting (the fly function gets added to the global scope). 

That's my two cents, but I'll leave it to you and Nicole to decide what goes where.

dgboss (approved):

PR #166, opened 2022-01-10
Removing code that incorrectly references other tools

michaelpnelson:
This comments out code that references other tools by using invalid tool names. Tool names were once all lowercase but are now camel case, but some references to the old tool names remain. This comments out such code because it is currently not functional. This may help prevent bugs, and by preserving the code we keep its intent should we want to bring back functionality later.

This also comments out nonfunctional buttons in the Search tool. The Identify and Measure buttons do not have implementation code, and the code called by the Directions button results in a console error because of an undefined value.

### DIFF
--- a/src/smk/tool/location/tool-location.js
+++ b/src/smk/tool/location/tool-location.js
@@ -31,20 +31,20 @@ include.module( 'tool-location', [
             this.geocoder = new SMK.TYPE.Geocoder( this.geocoderService )
 
             this.setIdentifyHandler = function ( handler ) {
-                if ( !smk.$tool.identify ) return
+                // if ( !smk.$tool.identify ) return
 
-                self.tool.identify = !!handler
+                // self.tool.identify = !!handler
 
-                self.identifyHandler = handler || function () {}
+                // self.identifyHandler = handler || function () {}
             }
             self.identifyHandler = function () {}
 
             this.setDirectionsHandler = function ( handler ) {
-                if ( !smk.$tool.directions ) return
+                // if ( !smk.$tool.directions ) return
 
-                self.tool.directions = !!handler
+                // self.tool.directions = !!handler
 
-                self.directionsHandler = handler || function () {}
+                // self.directionsHandler = handler || function () {}
             }
             self.directionsHandler = function () {}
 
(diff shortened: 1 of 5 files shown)


src/smk/tool/search/tool-search-location.js
```diff
@@ -42,26 +42,26 @@ include.module( 'tool-search.tool-search-location-js', [
...
-                                .catch( function ( err ) {
-                                    console.warn( err )
-                                    return smk.$tool.directions.addWaypoint()
-                                } )
-                        } )
-                }
-            } )
+            //         smk.$tool.directions.activating
+            //             .then( function () {
+            //                 return smk.$tool.directions.startAtCurrentLocation()
```
michaelpnelson: I spent a while trying to update references to "directions" with the new tools (it was replaced by three - DirectionsOptionsTool, DirectionsRouteTool, and DirectionsWaypointsTool, the last seeming to be the most appropriate) but I wasn't able to get it working.

src/smk/tool/location/tool-location.js
```diff
@@ -31,20 +31,20 @@ include.module( 'tool-location', [
             this.geocoder = new SMK.TYPE.Geocoder( this.geocoderService )
 
             this.setIdentifyHandler = function ( handler ) {
-                if ( !smk.$tool.identify ) return
+                // if ( !smk.$tool.identify ) return
```
dgboss: I'm not a fan of leaving commented out code in the code base, especially since Git will remember it for us. If you choose to go this route, could you please add comments to each section you are commenting out stating why you are doing it? Also, please consider commenting out code at a 'higher level' as well. For example, you might as well comment out the entire this.setIdentifyHandler function itself and comment out any calls to this function. Ditto for this.setDirectionsHandler.

src/smk/tool/location/tool-location.js
```diff
@@ -81,15 +81,15 @@ include.module( 'tool-location', [
...
-                            smk.$tool.directions.active = true
-
-                            smk.$tool.directions.activating
-                                .then( function () {
-                                    return smk.$tool.directions.startAtCurrentLocation()
-                                } )
-                                .then( function () {
-                                    return smk.$tool.directions.addWaypoint( site )
-                                } )
+                            // smk.$tool.directions.active = true
```
dgboss: Please add a comment indicating why this code is commented out.

src/smk/tool/search/tool-search-location.js
```diff
@@ -42,26 +42,26 @@ include.module( 'tool-search.tool-search-location-js', [
                 }
             } )
 
-            smk.on( this.id, {
-                'directions': function () {
-                    smk.$tool.directions.active = true
+            // smk.on( this.id, {
```
dgboss: Same here on the comment front please.

dgboss (changes requested):

src/smk/tool/location/tool-location.js
michaelpnelson: Thanks for your thoughts. I'm persuaded to remove the code instead of commenting it out. It's always more fun to remove unused code too :) . My only concern is being able to find this removed code some later day should we want to reinstate functionality ... much git tooling only shows the most recent change in a file/on a line, so it might be harder to see when it was removed. If you have suggestions, let me know. I think I'll update the commit message's first line to be very clear as to what was removed and why, since the first line is what we'll see in git history.

dgboss (approved):

PR #167, opened 2022-01-18
NPM package updates to address dependencies with vulnerabilities

michaelpnelson:


michaelpnelson:
The first commit fixes the "follow-redirects" vulnerability. I temporarily removed other dependencies that were causing failures when running 'npm audit fix', then ran 'npm audit fix', which updated package-lock.json.

The second commit adds "overrides" to package.json (https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides), allowing us to specify the version of a dependency's dependency. However, "overrides" require NPM version 8.3.

When testing locally, I found that I had to remove package-lock.json and delete my local node_modules directory for the overrides to take effect.

dgboss (approved):
Nice work!

PR #169, opened 2022-01-21
Update material-design-icons-iconfont version to match smk-cli

michaelpnelson:


michaelpnelson:
I noticed that, after this change, package-lock.json is updated with a material-design-icons-iconfont version of 6.1.1. However smk-cli has a version of 6.1.0 in its package-lock.json. This is a minor change but means that some icons like "percent" are missing in smk-cli. When I deleted and rebuilt package-lock.json in smk-cli, the version was 6.1.1.

I could push a PR updating package-lock.json to smk-cli. But what do you think of changing package.json in both to specify 6.1.1 or greater (from "^6.1.0" to ">=6.1.1"), instead?

dgboss:
To keep dependencies in sync, my recommendation would be to lock the version at 6.1.1. Should should consider doing this for all dependencies that are common to both projects.

michaelpnelson:
> To keep dependencies in sync, my recommendation would be to lock the version at 6.1.1. Should should consider doing this for all dependencies that are common to both projects.

Sounds good. I'll update the following common dependencies to be versions only, here and for smk-cli:

- "glob": "7.1.2"
- "grunt": 1.4.1
- "grunt-contrib-watch": "1.1.0"
- "load-grunt-tasks": "4.0.0"
- "material-design-icons-iconfont": "6.1.1"

UPDATE: Freezing "grunt" at 1.0.3 introduced 4 high priority audit warnings. Following the recommendation, it will be updated to 1.4.1.

michaelpnelson:
Related PR in smk-cli: https://github.com/bcgov/smk-cli/pull/87

dgboss (approved):
Looks good.

PR #170, opened 2022-02-02
Vector tile layer demo

michaelpnelson:
This adds a demo of vector tile layers using protocol buffers and sliced JSON.


debug/customization/vector-tile-layer/index.html
```diff
@@ -0,0 +1,24 @@
...
+    </head>
+
+    <body>
+        <header>Vector Tile Layer Demo</header>
+
+        <article>
+            <div id="smk-map-frame"></div>
+        </article>
+
+        <footer>Application constructed by smk-cli at 2022-01-24T19:49:23.423Z</footer>
```
dgboss: I don't think this is necessary.

debug/customization/vector-tile-layer/smk-config.json
```diff
@@ -0,0 +1,124 @@
...
+            "type": "geomark",
+            "geomarkService": {
+                "url": "https://apps.gov.bc.ca/pub/geomark"
+            },
+            "position": "list-menu",
+            "enabled": false
+        }
+    ],
+    "version": "1.0.8",
+    "createdBy": "Michael Nelson",
```
dgboss: I'd suggest using either SMK Developer or MAP team or something else generic, but it's up to you.

debug/customization/vector-tile-layer/smk-init.js
```diff
@@ -0,0 +1,298 @@
...
+                vectorTileStyling.buildings  = vectorTileStyling.building;
+                vectorTileStyling.boundaries = vectorTileStyling.boundary;
+                vectorTileStyling.places     = vectorTileStyling.place;
+                vectorTileStyling.pois       = vectorTileStyling.poi;
+                vectorTileStyling.roads      = vectorTileStyling.road;
+
+            	// Nextzen tiles PBF layer setup
+
+                // Assumes layers = "all", and format = "mvt"
+                var nextzenTilesUrl = "https://tile.nextzen.org/tilezen/vector/v1/512/all/{z}/{x}/{y}.mvt?api_key={apikey}";
```
dgboss: We shouldn't be using var.

debug/customization/vector-tile-layer/smk-init.js
```diff
@@ -0,0 +1,298 @@
...
+                        fill: true,
+                        weight: 1,
+                        fillColor: '#c0c0c0',
+                        color: '#c0c0c0',
+                        fillOpacity: 0.2,
+                        opacity: 0.4
+                    }
+                };
+
+                // Monkey-patch some properties for nextzen layer names, because
```
dgboss: Is this necessary? It looks like the copied code was last updated in 2019.

debug/customization/vector-tile-layer/smk-init.js
```diff
@@ -0,0 +1,298 @@
...
+            	// ESRI tiles PBF layer setup
+                var esriStyle = {};
+                esriStyle.Continent  = vectorTileStyling.earth;
+                esriStyle.Bathymetry = vectorTileStyling.water;
+                esriStyle["Vegetation small scale"] = vectorTileStyling.landuse;
+                esriStyle["Marine area"] = vectorTileStyling.water;
+                esriStyle.Land = vectorTileStyling.earth;
+
+                esriStyle["City small scale"             ] = vectorTileStyling.building;
+                esriStyle["Admin0 point"                 ] = [];
```
dgboss: Does this actually need to be defined as an empty array?

debug/customization/vector-tile-layer/smk-init.js
```diff
@@ -0,0 +1,298 @@
...
+					return f.properties.wb_a3;
+				}
+            }
+		);
+
+		// Set an extent that includes BC as well as Europe
+		leafletMap.setView({ lat: 48, lng: -40 }, 3);
+
+		// Add toggling of layers
+		L.control.layers({}, {
```
dgboss, line 68: Have you looked to see if you can get the layers to display in the SMK layer control?

debug/customization/vector-tile-layer/smk-config.json
```diff
@@ -0,0 +1,124 @@
...
+            "enabled": true
+        },
+        {
+            "type": "zoom",
+            "enabled": true
+        },
+        {
+            "type": "about",
+            "enabled": true,
+            "title": "About Vector Layer Demo!",
```
dgboss: If we're going to have the About tool enabled, we should add a helpful description.

debug/customization/vector-tile-layer/smk-init.js
```diff
@@ -0,0 +1,298 @@
...
+            	// Nextzen tiles PBF layer setup
+
+                // Assumes layers = "all", and format = "mvt"
+                var nextzenTilesUrl = "https://tile.nextzen.org/tilezen/vector/v1/512/all/{z}/{x}/{y}.mvt?api_key={apikey}";
+
+                var nextzenVectorTileOptions = {
+                    rendererFactory: L.canvas.tile,
+                    attribution: '<a href="https://nextzen.com/">&copy; NextZen</a>, <a href="http://www.openstreetmap.org/copyright">&copy; OpenStreetMap</a> contributors',
+                    vectorTileLayerStyles: vectorTileStyling,
+                    apikey: 'gCZXZglvRQa6sB2z7JzL1w',
```
dgboss: Where did this key come from? Is it your personal key? Should we be including it in our repo?

dgboss (changes requested):

debug/customization/vector-tile-layer/index.html
michaelpnelson: The header? I will remove it.

debug/customization/vector-tile-layer/smk-config.json
michaelpnelson: I'll update this to use "SMK Developer".

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: Thanks. I've replaced vars that are never modified as consts, and vars that are scoped to a block as lets.

debug/customization/vector-tile-layer/index.html
dgboss: Oops, comment didn't insert correctly. The footer.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: I'm not sure how we'd confirm whether the nextzen properties being renamed (buildings, boundaries, etc.) are still in use - any suggestions?

debug/customization/vector-tile-layer/smk-init.js
dgboss: Choosing between let and const actually shouldn't depend on the scope; let should only be used for variables that have a value re-assigned to them and const for everything else.

debug/customization/vector-tile-layer/smk-init.js
dgboss: Looking at how the vectorTileStyling object is used, you can simply use the correct property names when the object is first declared. For example, 
```
var vectorTileStyling = {
    buildings: { 
        fill: true,
        weight: 1,
        fillColor: '#06cccc',
        color: '#06cccc',
        fillOpacity: 0.2,
        opacity: 0.4,
    }, ...
```
Notice that I've replaced building with buildings, there will no longer be a buildings property. You will then have to update the esriStyle object, 
```
esriStyle["City small scale"] = vectorTileStyling.buildings;
```

michaelpnelson: Doing this seems to step on some Esri style names. For example, it looks like Esri has both "boundary" and "boundaries", so the replacement means the internal boundaries between states and provinces go from the earlier light magenta to a thicker blue.

I vote to leave this as-is. Earlier a comment makes it clear the code is copied, and gives the source URL.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: It looks to be useful. For example, if I remove the empty arrays for "water line" style names, rivers are given a thick blue style, which looks distracting.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: I spent some time attempting to but didn't find a way to do so.

If we add proper support for vector tile layers, much of the effort will be in getting our layer tool to work with them.

debug/customization/vector-tile-layer/smk-config.json
michaelpnelson: Er, yes, I forgot to update this after demoing this internally ... I'll add something more appropriate :)

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: I'll go ahead and disable the layer tool since it's ineffective with these layers.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: This key was copied from the Leaflet.VectorGrid sample code (https://leaflet.github.io/Leaflet.VectorGrid/demo-vectortiles.html). This source is acknowledged in an earlier comment.

debug/customization/vector-tile-layer/smk-init.js
dgboss: I think every instance of 'let' in this file can actually be replaced with 'const'.

michaelpnelson: I confirmed we can replace all 'let's with 'const's since we don't reassign values to any of the 'let's - we only update objects or arrays.

debug/customization/vector-tile-layer/smk-init.js
```diff
@@ -0,0 +1,298 @@
...
+
+    include([
+		// Load Leaflet VectorGrid
+        {
+            url: "https://unpkg.com/leaflet.vectorgrid@1.2.0",
+            loader: "script"
+        },
+		// Load EU countries JSON as 'euCountries'
+        {
+            url: "https://leaflet.github.io/Leaflet.VectorGrid/eu-countries.js",
```
dgboss: Can this eu-contries file be sourced from a CDN somewhere? If not, we should download a local copy and use it; otherwise, if this file becomes unavailable the whole demo breaks.

debug/customization/vector-tile-layer/smk-init.js
dgboss: I'm not seeing any issues when I make the suggested change, boundaries continue to appear as magenta provided you update the associated property in the esriStyle object. 

The reason I commented on this is that I feel that leaving this 'monkey patch' in place adds unnecessary complexity and is confusing. I feel it's not an example of good coding practice and don't really want to demonstrate this in our code base. I'll leave the final decision up to you.

debug/customization/vector-tile-layer/smk-init.js
dgboss: Could you add a code comment at the top of the declaration of esriStyle that explains what the empty arrays do? ie. They remove the styling for that object/proprty in the vector tile so they don't appear on the map.

debug/customization/vector-tile-layer/smk-init.js
```diff
@@ -0,0 +1,298 @@
...
+					}
+				},
+				getFeatureId: function(f) {
+					return f.properties.wb_a3;
+				}
+            }
+		);
+
+		// Set an extent that includes BC as well as Europe
+		leafletMap.setView({ lat: 48, lng: -40 }, 3);
```
dgboss: Consider including the initial extent in the smk-config.json instead. It looks a little strange for the map to start at the extent of BC and then zoom out.

debug/customization/vector-tile-layer/smk-init.js
dgboss: I don't think we should be using or promoting the use of an api key we don't own. Does nextzen have any free layers? If not, we might want to replace this with a different example.

dgboss:
What do you think about having the style objects defined in their own file, ie. move vectorTileStyling and esriStyle into a separate file? Just a thought to clean things up a little, though there are pros and cons...

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: I didn't find it in a CDN so I'll download it. I'll add it in a <script> in index.html; adding it here as "eu-countries.js" results in smk.js looking for it in /dist/assets/src/eu-countries.js, not finding it, and giving an error.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: You're right, I'd probably forgotten to update the Esri style as "boundaries" - I can reproduce what I saw earlier by not updating the Esri "boundary" as "boundaries". 

I'll reapply my earlier changes and address the omission.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: Done.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: I'm leaning towards removing the Nextzen layer then - what do you think? The Leaflet protobuf examples (https://leaflet.github.io/Leaflet.VectorGrid/demo-vectortiles.html) all use a key of some kind except for Esri.

debug/customization/vector-tile-layer/smk-init.js
dgboss: I can load it like this on a Windows machine after placing a copy in the assets directory:

```
    {
            url: "./assets/euCountries.js",
            loader: "script"
    }
```

debug/customization/vector-tile-layer/smk-init.js
dgboss: Yeah, we should probably remove the Nextzen layer the then.

michaelpnelson: I'll remove the Nextzen layer, and any layer styles that aren't used in the Esri layer.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: Done.

debug/customization/vector-tile-layer/smk-init.js
michaelpnelson: This works for me too. Done.

michaelpnelson:
> What do you think about having the style objects defined in their own file, ie. move vectorTileStyling and esriStyle into a separate file? Just a thought to clean things up a little, though there are pros and cons...

I'm fine moving the style setup to its own file. The code in smk-init.js becomes much more focused in its absence. This will be done in the next commit.

debug/customization/vector-tile-layer/smk-config.json
```diff
@@ -0,0 +1,124 @@
...
+        {
+            "type": "zoom",
+            "enabled": true
+        },
+        {
+            "type": "about",
+            "enabled": true,
+            "title": "About",
+            "position": "list-menu",
+            "content": "<p>This demonstrates the use of vector tile layers of multiple types (<a href=\"https://docs.mapbox.com/data/tilesets/guides/vector-tiles-standards/\">Protobufs</a> and sliced <a href=\"https://geojson.org/\">GeoJSON</a>) in SMK using the <a href=\"https://github.com/Leaflet/Leaflet.VectorGrid\">Leaflet.VectorGrid</a> library.</p><p>While not yet supported in the Layers tool, vector tile layers can be added to SMK apps with custom javascript.</p>"
```
dgboss: javascript should be JavaScript.

dgboss (approved):

debug/customization/vector-tile-layer/smk-config.json
michaelpnelson: Done.

PR #172, opened 2022-02-18
NPM package updates addressing vulnerabilities

michaelpnelson:
This updates our dependencies on two modules as well as adding an override for a transitive dependency. These changes result in zero vulnerabilities when running 'npm audit'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
   },
   "dependencies": {},
   "devDependencies": {
-    "browser-sync": "^2.27.5",
-    "concurrently": "^5.3.0",
+    "browser-sync": "^2.27.7",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "material-design-icons-iconfont": "6.1.1"
   },
   "overrides": {
+    "engine.io": ">=6.1.2",
```
dgboss, line 40: Same issue as the smk-cli PR, the overrides don't seem to be working?

dgboss (changes requested):

dgboss (approved):

michaelpnelson:
Screen capture of the new override version applied to node_modules...

smk-engine-io

PR #173, opened 2022-02-24
Fix issue loading multipolygon Geomarks

michaelpnelson:


src/smk/tool/geomark/tool-geomark.js
```diff
@@ -211,14 +211,45 @@ include.module( 'tool-geomark', [
             }
 
             this.toGeomark = function(geomarkInfo, drawingLayer) {
+                const geomarkProperties = this.getGeomarkProperties(geomarkInfo);
                 return {
-                    id: geomarkInfo.id || geomarkInfo.properties.id,
-                    url: geomarkInfo.url || geomarkInfo.properties.url,
+                    id: geomarkInfo.id || geomarkProperties.id,
```
dgboss: It looks like geomarkInfo.id and geomarkInfo.url are always going to be undefined.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -211,14 +211,45 @@ include.module( 'tool-geomark', [
...
                 };
             }
 
+            this.getGeomarkProperties = function(geomarkInfo) {
+                // polygon
+                if (geomarkInfo.properties) {
+                    return geomarkInfo.properties;
+                }
+                // multi-polygon
+                if (geomarkInfo.features && geomarkInfo.features[0] && geomarkInfo.features[0].properties) {
```
dgboss, line 234: I'm having a forgetful moment, will indexing into an empty array like this throw an exception in JavaScript? I might just be thinking about other languages (like C#).

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -243,23 +274,24 @@ include.module( 'tool-geomark', [
                     url: geomarkUrl + '/feature.geojson',
                     dataType: 'json',
                     traditional: true,
-                    success: function(geomarkFeature) {
-                        var geometryLayer = L.geoJSON(geomarkFeature.geometry, {
+                    success: function(geomarkGeoJson) {
+                        const geometry = self.getGeometry(geomarkGeoJson);
+                        const geometryLayer = L.geoJSON(geometry, {
```
dgboss: An L.geoJson layer can accept an input parameter of geojson, so there is no need to parse out a geometry here, you can just do
```
const geojsonLayer = L.geoJson(geomarkGeoJson, {
  style: function(feature) {
    return {
      color: CUSTOM_COLOR
    };
  }
);
```

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -211,14 +211,45 @@ include.module( 'tool-geomark', [
...
+                }
+                // multi-polygon
+                if (geomarkInfo.features && geomarkInfo.features[0] && geomarkInfo.features[0].properties) {
+                    return geomarkInfo.features[0].properties;
+                } 
+                self.showCouldNotLoadMessage();
+                return {};
+            }
+
+            this.getGeometry = function(geomarkGeoJson) {
```
dgboss: I don't think this getGeometry function is necessary (see comment below about L.geoJSON layer.

src/smk/tool/geomark/tool-geomark.js
dgboss: The getGeomarkProperties function has a path where it returns an empty object. If this happens, accessing a property of an empty object will throw an unhandled exception (eg. geomarkProperties.id and geomarkProperties.url).

dgboss (changes requested):

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: They'll be undefined now and going forward since Geomarks API isn't expected to revert. However at some point these must have had a value (I can't imagine why we'd have added it otherwise). 

I don't yet know whether a Geomark created in an earlier API version will give the same response on GET in a newer API version as it did for the earlier API version. I suspect the response will always follow the current API regardless of the API version it was created with but if it doesn't then keeping this backwards compatibility will be important.

Do you mind if we keep this as-is until we can confirm Geomark API behaviour on this?

michaelpnelson: I returned the empty object because I know that attempting to access a property of an undefined object would throw an exception ... but if accessing a property on an empty object ALSO throws an exception, returning an empty object is no better. 

The only other thought I have is to throw an exception - how does that sound? I'll try it out, and move the call to showCouldNotLoadMessage() into the catch block. Same for the getGeometry() function, which can also return an empty object.

dgboss: No, you're right. The property values will just be undefined, I'm getting my languages mixed up! But is a geomark with an undefined id and/or url valid?

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Thanks! I've confirmed we can successfully make layers from the received GeoJSON (polygons and multipolygons) using prod, delivery, and test Geomark environments and their varied JSON structures.

I'm happy to rely on Leaflet's parsing and not have to maintain our own :)

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Removed

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Update: We also use this method (toGeomark) when creating a Geomark, and in that case, geomarkInfo.id and geomarkInfo.url have a value.

michaelpnelson: Another update: I moved the call to showCouldNotLoadMessage() to the last part of these "chained ORs". I saw that when creating a Geomark, we'd call getGeomarkProperties() and the properties would not be found in the deeper structures where we look for them (because they're at the root in that situation) and we'd incorrectly show the message.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -211,14 +211,31 @@ include.module( 'tool-geomark', [
             }
 
             this.toGeomark = function(geomarkInfo, drawingLayer) {
+                const geomarkProperties = this.getGeomarkProperties(geomarkInfo);
                 return {
-                    id: geomarkInfo.id || geomarkInfo.properties.id,
-                    url: geomarkInfo.url || geomarkInfo.properties.url,
+                    id: geomarkInfo.id || geomarkProperties.id || this.showCouldNotLoadMessage(),
```
dgboss: I understand that you want to show an error message when a Geomark can't be created with an id or url. The `showCouldNotLoadMessage` function will return undefined by default, resulting in your Geomark object having an undefined value for id, url or both. Is this your intention? Or should `toGeomark` be returning undefined if id/url is undefined?

michaelpnelson: I didn't consider the return value of calling showCouldNotLoadMessage(). That said, returning undefined values after displaying the message is no worse than returning the undefined values we get from the args, and at least a message is displayed. 

Alternately, we could remove the call to showCouldNotLoadMessage() altogether, since we currently don't expect those values to be undefined after the new changes - maybe it's overkill.

I vote to leave this as-is so that a message is displayed if the Geomark API is changed again. What do you think?

dgboss: Sorry, I'm not explaining myself very well. You're right in that it doesn't matter if the `undefined` value being assigned to the id or url is coming from the return value of the function, or `geomarkInfo.id`, 'geomarkInfo.url` etc...

I may be incorrect, but from my point of view, what matters is that the `toGeomark` function is returning an object that would appear to be in an invalid state with undefined `id` and/or `url` properties. The callers of the `toGeomark` function then add this invalid (again, my opinion) geomark to the `self.geomark` collection.

My suggestion would be something like this:
```
this.toGeomark = function(geomarkInfo, drawingLayer) {
    const geomarkProperties = this.getGeomarkProperties(geomarkInfo);
    const id = geomarkInfo.id || geomarkProperties.id;
    const url = geomarkInfo.url || geomarkProperties.url;

    if (id && url) {
      return {
          id,
          url,
          drawingLayer: drawingLayer,
          isVisible: true
     };

     this.showCouldNotLoadMessage();
     return undefined; 
}
```

Then at lines 271 and 304 where `toGeomark` is called, I would check if the value being returned is undefined and act accordingly.

Alternatively, if you're confident that there will always be id and url values available, simply remove your call to `showCouldNotLoadMessage` and ignore me!

michaelpnelson: Thanks for your suggestion.

My first thought was to update both callers to check for undefined values and handle undefined values by calling showCouldNotLoadMessage(). But on further thought, I like your suggestion that toGeomark() return either a response without any undefined properties, or "undefined" - either/or. Then calling code can just add an "if" block before proceeding. So I think I'll go with that.

dgboss (approved):

PR #176, opened 2022-04-19
ESRI layer docs mention drawingInfo and showLabels

michaelpnelson:
This updates documentation for ESRI dynamic layers by adding reference for the 'drawingInfo' property. It also mentions that SMK-CLI can toggle the display of labels via the 'showLabels' property.

### DIFF
--- a/docs/configuration/layers/esri-dynamic.md
+++ b/docs/configuration/layers/esri-dynamic.md
@@ -61,9 +61,11 @@ The URL for the DataBC Layer Catalog.
 ## DynamicLayers Property
 `"dynamicLayers": Array`
 
-A listing of dynamic layer configurations.
-This will typically only contain one dynamic feature, which can be derived from the MPCM Layer Catalog.
+A listing of dynamic layer configurations in JSON format. This will typically only contain one dynamic feature. This configuration comes from the MPCM Layer Catalog.
 
+### DrawingInfo Property
+`"drawingInfo": Object`
 
+Each `dynamicLayers` property includes a `drawingInfo` property. This is an object in JSON format that is included in SMK requests for layer tiles. `drawingInfo` can be manually edited to configure aspects of layer display. For more information, see <a href="https://developers.arcgis.com/web-map-specification/objects/drawingInfo">`drawingInfo` documentation</a>.
 
-
+If a value exists for `drawingInfo`'s `labelingInfo` property in an ESRI dynamic layer, SMK-CLI's panel for editing the layer will include a "Show labels" check box. This toggles the value of the `showLabels` property of `drawingInfo`. 
--- a/docs/configuration/layers/esri-dynamic.md
+++ b/docs/configuration/layers/esri-dynamic.md
@@ -61,11 +61,9 @@ The URL for the DataBC Layer Catalog.
 ## DynamicLayers Property
 `"dynamicLayers": Array`
 
-A listing of dynamic layer configurations in JSON format. This will typically only contain one dynamic feature. This configuration comes from the MPCM Layer Catalog.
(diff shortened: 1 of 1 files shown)


docs/configuration/layers/esri-dynamic.md
```diff
@@ -61,9 +61,11 @@ The URL for the DataBC Layer Catalog.
 ## DynamicLayers Property
 `"dynamicLayers": Array`
 
-A listing of dynamic layer configurations.
-This will typically only contain one dynamic feature, which can be derived from the MPCM Layer Catalog.
-
+A listing of dynamic layer configurations in JSON format. This will typically only contain one dynamic feature. This configuration comes from the MPCM Layer Catalog.
```
dgboss: I would suggest leaving out the 'in JSON format' here and further down.

dgboss: I know it already existed, but 'dynamic feature' doesn't make sense. I'd suggest something like 'This will typically contain configuration for a single layer. The default configuration comes from the MPCM Layer Catalog.'

docs/configuration/layers/esri-dynamic.md
```diff
@@ -61,9 +61,11 @@ The URL for the DataBC Layer Catalog.
...
 
+### DrawingInfo Property
+`"drawingInfo": Object`
 
+Each `dynamicLayers` property includes a `drawingInfo` property. This is an object in JSON format that is included in SMK requests for layer tiles. `drawingInfo` can be manually edited to configure aspects of layer display. For more information, see <a href="https://developers.arcgis.com/web-map-specification/objects/drawingInfo">`drawingInfo` documentation</a>.
```
dgboss: We're not actually requesting tiles, but images rendered dynamically on the server. Consider...
"Each `dynamicLayers` element includes a `drawingInfo` property which defines the appearance of the layer on the map."

docs/configuration/layers/esri-dynamic.md
```diff
@@ -61,9 +61,11 @@ The URL for the DataBC Layer Catalog.
...
 
+If a value exists for `drawingInfo`'s `labelingInfo` property in an ESRI dynamic layer, SMK-CLI's panel for editing the layer will include a "Show labels" check box. This toggles the value of the `showLabels` property of `drawingInfo`.
```
dgboss: I'm not sure we should be referencing the SMK-CLI here in the SMK project, though this would fit very well in the SMK-CLI docs. We could say that `drawingInfo.labelingInfo` defines the appearance of labels for this layer. We could also mention the `showLabels` property somewhere?

dgboss (changes requested):

docs/configuration/layers/esri-dynamic.md
michaelpnelson: I guess it _is_ pretty obvious. Done.

michaelpnelson: I'm fine with your suggestion. Done.

docs/configuration/layers/esri-dynamic.md
michaelpnelson: Thanks for clarifying. I'm fine with your suggestion.

docs/configuration/layers/esri-dynamic.md
michaelpnelson: I'll remove mention of SMK-CLI here and make a new PR to update https://bcgov.github.io/smk-cli/edit-layer.html.

Given that showLabels will be documented in smk-cli, I will just remove this paragraph altogether. I don't think there's a need to call out labelingInfo and showLabels here - they're just parts of drawingInfo among the others, and we already link to ESRI docs for drawingInfo.

dgboss (approved):

PR #177, opened 2022-04-19
Update 'async' version to address vulnerability

michaelpnelson:
This updates the 'async' dependency's version to require a version that fixes a recent security vulnerability. 'grunt-contrib-clean' is also updated as it uses async but also has a substantial update in version. package-lock.json is also updated after deleting and reinstalling node modules.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "glob": "7.1.2",
         "grunt": "1.4.1",
         "grunt-cli": "~1.4.0",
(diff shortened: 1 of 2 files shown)


dgboss (changes requested):
These changes break the npm run start command. I'm not sure what the underlying issue is. To test:
1. Delete your node_modules directory
2. npm ci
3. npm run start

dgboss (approved):

PR #179, opened 2022-05-17
Version incremented, redundant font-family style removed

michaelpnelson:
This increments the version to synchronize with related changes in SMK-CLI where the BC Sans font will be added and set in the main style sheet for created SMK apps. SMK-CLI changes will depend on this incremented version.

This also removes the font-family style in the map frame styling, which has the same value as the font-family style in style.css (added to new applications by 'smk create'). We remove it because it is redundant and because it overrides the font-family change to BC Sans we want to make in smk-cli to style.css.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@bcgov/smk",
-  "version": "1.0.9",
+  "version": "1.0.10",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


michaelpnelson:
The related smk-cli changes are in branch https://github.com/bcgov/smk-cli/tree/bcsans_newapps. I'm planning to create a pull request for it (plus changes to require smk version 1.1.10+) after these changes are integrated.

dgboss (approved):
Looks good to me.

PR #181, opened 2022-07-08
Basemaps updated for new ESRI API

michaelpnelson:
This updates basemap code for ESRI basemaps to use the esri-leaflet-vector library and accommodate the use of an API key. Docs are updated, debug examples are updated to use a non-ESRI basemap, and SMK's version is incremented.


michaelpnelson:
This is a big one! Please take your time. If you'd like to see the changes in smaller commits, I can push a local branch that has the same changes un-squashed.

I'm pushing this to a feature branch so that we can delay integration with master until we're ready.

I'll be working on basemap updates in smk-cli next. Much has already been done in branch "wip_basemaps_update". Feel free to check that out if you'd like to use it to make config changes with the new basemaps.

src/smk/viewer-leaflet/viewer-leaflet.js
```diff
@@ -29,6 +29,9 @@ include.module( 'viewer-leaflet', [ 'viewer', 'leaflet', 'layer-leaflet', /*'fea
             minZoom:            smk.viewer.minZoom
         } )
 
+        self.map.createPane('basemaps');
```
dgboss: Could you please add a comment explaining what a 'pane' is and why this is necessary (for future reference).

src/smk/viewer-leaflet/viewer-leaflet.js
```diff
@@ -177,47 +180,46 @@ include.module( 'viewer-leaflet', [ 'viewer', 'leaflet', 'layer-leaflet', /*'fea
...
-    function createBasemapEsri( id ) {
+    function createBasemapEsri( esriApiKey, apiId ) {
+        if (!esriApiKey) {
+            throw new Error('No value was found for "esriApiKey" while creating an ESRI vector basemap. As a result, no basemap will appear. Please confirm there is a value for "esriApiKey" in the "viewer" section of smk-config.json, or use a non-ESRI basemap.');
+        }
+
         /* jshint -W040 */
-        var opt = Object.assign( { detectRetina: true }, this.option )
+        var opt = Object.assign( { detectRetina: true }, this.option );
+        opt.pane = 'basemaps';
```
dgboss: Rather than using a hard-coded string here (and above at lines 32 and 33 and below at line 219), consider defining a constant that can be used at each spot.

src/smk/viewer-leaflet/viewer-leaflet.js
```diff
@@ -177,47 +180,46 @@ include.module( 'viewer-leaflet', [ 'viewer', 'leaflet', 'layer-leaflet', /*'fea
...
+        opt.apikey = esriApiKey;
 
         var lys = []
-        lys.push( L.esri.basemapLayer( id, opt ) )
+        lys.push( L.esri.Vector.vectorBasemapLayer( apiId, opt ) )
 
         if ( this.labels )
```
dgboss: Can this block be removed? I don't think we use label layers anymore?

src/smk/viewer-leaflet/viewer-leaflet.js
```diff
@@ -177,47 +180,46 @@ include.module( 'viewer-leaflet', [ 'viewer', 'leaflet', 'layer-leaflet', /*'fea
...
     }
 
     function createBasemapTiled( id ) {
         /* jshint -W040 */
-        return [ L.tileLayer( this.url, { attribution: this.attribution } ) ]
+        return [ L.tileLayer( this.url, { attribution: this.attribution, pane: 'basemaps' } ) ]
     }
 
-    ViewerLeaflet.prototype.setBasemap = function ( basemapId ) {
+    ViewerLeaflet.prototype.setBasemap = function ( esriApiKey, basemapId ) {
```
dgboss: When adding a new parameter(s) to a function, it's generally better to add them after existing parameters.

src/smk/viewer-leaflet/viewer-leaflet.js
```diff
@@ -226,19 +228,19 @@ include.module( 'viewer-leaflet', [ 'viewer', 'leaflet', 'layer-leaflet', /*'fea
...
             this.map.addLayer( this.currentBasemap[ i ] );
 
         this.changedBaseMap( { baseMap: basemapId } )
     }
 
-    ViewerLeaflet.prototype.createBasemapLayer = function ( basemapId ) {
-        return this.basemap[ basemapId ].create( basemapId )
+    ViewerLeaflet.prototype.createBasemapLayer = function ( esriApiKey, basemapId ) {
+        const basemap = this.basemap[basemapId];
+        return basemap.create( esriApiKey, basemap.apiId || basemap.id )
```
dgboss: Is there a reason we need both of these options (basemap.apiId and basemap.id)?

src/smk/viewer-leaflet/tool/minimap/tool-minimap-leaflet.js
```diff
@@ -6,7 +6,7 @@ include.module( 'tool-minimap-leaflet', [ 'leaflet', 'tool-minimap' ], function
 
         smk.addToStatus( $( '<div class="smk-spacer">' ).height( 170 ).get( 0 ) )
 
-        var ly = smk.$viewer.createBasemapLayer( this.baseMap || "Topographic" );
+        var ly = smk.$viewer.createBasemapLayer( smk.viewer.esriApiKey, this.baseMap || "Topographic" );
```
dgboss: Should we make the default StamenTonerLight?

src/smk/tool/baseMaps/tool-baseMaps.js
```diff
@@ -57,7 +57,10 @@ include.module( 'tool-baseMaps', [
                             zoomSnap: 0
                         } );
 
-                        var bmLayers = smk.$viewer.createBasemapLayer( bm.id )
+                        m.createPane('basemaps');
+                        m.getPane('basemaps').style.zIndex = 100;
+
+                        var bmLayers = smk.$viewer.createBasemapLayer( smk.viewer.esriApiKey, bm.id )
```
dgboss: I'm not sure we need to read the esriApiKey value from the smk.viewer config, only to pass it back to the viewer object. It might be a little cleaner/maintainable if we only access the key where it is needed (ie. only within `createBasemapEsri` in viewer-leaflet.js).

src/smk/viewer-leaflet/tool/minimap/tool-minimap-leaflet.js
dgboss: Same idea here as for the basemap tool, I don't think the minimap tool needs to be aware of the esriApiKey.

dgboss (changes requested):
Nice work! This is working great!

dgboss:
One more thought, did we want to make StamenTonerLight the default basemap? If so, we wouldn't need to make all those changes to the demo config files.

src/smk/tool/baseMaps/tool-baseMaps.js
michaelpnelson: The reason it's passed around is that we only have a reference to the "smk" object in the init method, and then pass it from init to where it's needed. 

We might be able to do this using the SMK global variable. I see there's SMK.CONFIG.Viewer, which includes the "basemap" property, but it looks like I'd need to add the "esriApiKey" property. Is this what you were thinking, or do you know of a better way to get the property from within the createBasemapEsri() method?

michaelpnelson: I looked into this further ... it looks like the values in SMK.CONFIG don't get updated as the app is initialized and run; they remain as set in smk.js. For example, SMK.CONFIG.viewer.baseMap is set as "Topographic" and shows that value when debugging, regardless of the basemap set in smk-config.js.

michaelpnelson:
> One more thought, did we want to make StamenTonerLight the default basemap? If so, we wouldn't need to make all those changes to the demo config files.

I'd rather keep Topographic as the default basemap. I think it's a good all-rounder choice. Plus, I've already done all the work with changing the demo files :) .

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: Done

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: Done. I define the const high in the class (outside of where it's first used in ViewerLeaflet.prototype.initialize) so that it is visible in other functions.

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: I confirmed that we don't use label layers and removed this block.

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: Done.

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: Yes - the basemapId used to always be the same value we'd use when calling API to create the basemap. With this change, a basemapId like "Topographic" can't be used with the new Esri API - instead we have to use new value "ArcGIS:Topographic".

src/smk/viewer-leaflet/tool/minimap/tool-minimap-leaflet.js
michaelpnelson: I think we should keep Topographic the default - see my reply on the review comment.

src/smk/viewer-leaflet/viewer-leaflet.js
dgboss: When I inspect the this.basemap array, none of the items have an `id` property. I think you can get away with just keeping the basemap.apiId.

src/smk/viewer-leaflet/tool/minimap/tool-minimap-leaflet.js
dgboss: My concern with using an Esri basemap is that apps appear broken if you haven't changed the basemap to StamenTonerLight or added configured an esriApiKey. This doesn't make for a great first experience for new users. Perhaps worth discussing with Nicole?

src/smk/tool/baseMaps/tool-baseMaps.js
dgboss: The config is available on the global SMK object at `SMK.MAP[0].viewer`. The reason I mention it is that we end up passing around the key when it is not needed (eg. for the StamenTonerLight basemap). Though I suppose we also pass an unused id into `basemap.create()` when the basemap is StamenTonerLight. I'll leave the choice up to you.

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: We still need id for StamenTonerLight. The value of this.basemap is defined in viewer.js, and in this review the apiId is added for all basemaps except StamenTonerLight. The id is intrinsic (which I learned while working on this :) ) and takes the value of the key, 'StamenTonerLight'.

src/smk/viewer-leaflet/tool/minimap/tool-minimap-leaflet.js
michaelpnelson: Nicole and I discussed upgrade scenarios and options. As to whether to use StamenTonerLight as a default basemap (both in SMK-CLI, and in internal code when there's no basemap in config) ... we've decided to leave it as Topographic because if we did change to StamenTonerLight, it would only address new SMK apps, and we would most want to address legacy SMK apps.

src/smk/tool/baseMaps/tool-baseMaps.js
michaelpnelson: Thanks for suggesting SMK.MAP, though when I debug I only see one member: SMK.MAP[1] - if this seems incorrect please let me know. I prefer to reference this global value than pass around a value and will update for this.

src/smk/viewer-leaflet/tool/minimap/tool-minimap-leaflet.js
dgboss: Ok, I still don't like the idea of delivering a product that appears broken out of the box, but I'll stand aside and protest silently!

src/smk/viewer-leaflet/viewer-leaflet.js
dgboss: To satisfy my curiosity...what is an intrinsic id? Is this something unique to Vue? In this case, basemap.id always seems to be undefined for me, even in the `createBasemapTiled()` function.

dgboss (approved):

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: My apologies ... I'd earlier come to an incorrect conclusion. I also see that basemap.id is always undefined. I accepted the "intrinsic id" explanation because this change always worked for StamenTonerLight, but now I notice that code previously used "basemapId" and I replaced it with "basemap.id". But the reason it works is that the "id" argument in createBasemapTiled(id) is not used; in that case we create an L.tileLayer with the url and an object of other parameters.

In the next change set I'll restore "basemap.id" as "basemapId".

PR #182, opened 2022-07-20
Basemaps updated for new ESRI API

michaelpnelson:
This updates basemap code for ESRI basemaps to use the esri-leaflet-vector library and accommodate the use of an API key. Docs are updated, debug examples are updated to use a non-ESRI basemap, and SMK's version is incremented.


src/smk/viewer-leaflet/viewer-leaflet.js
```diff
@@ -177,44 +183,38 @@ include.module( 'viewer-leaflet', [ 'viewer', 'leaflet', 'layer-leaflet', /*'fea
...
     ViewerLeaflet.prototype.basemap.StamenTonerLight.attribution = "Map tiles by <a href='http://stamen.com'>Stamen Design</a>, under <a href='http://creativecommons.org/licenses/by/3.0'>CC BY 3.0</a>. Data by <a href='http://openstreetmap.org'>OpenStreetMap</a>, under <a href='http://www.openstreetmap.org/copyright'>ODbL</a>."
 
-    function createBasemapEsri( id ) {
-        /* jshint -W040 */
-        var opt = Object.assign( { detectRetina: true }, this.option )
-
-        var lys = []
-        lys.push( L.esri.basemapLayer( id, opt ) )
+    function createBasemapEsri( apiId ) {
+        const apiKey = SMK.MAP[1].viewer.esriApiKey;
```
dgboss: This lookup is unsafe, SMK.MAP[1] doesn't always exist. The '1' is the 'smk-id' and can be set by the user. See the SMK/Initialization docs for details.

dgboss (changes requested):

src/smk/viewer-leaflet/viewer-leaflet.js
michaelpnelson: I'll reinstate the passing around of the "esriApiKey" value that we get from the smk object in the initialization method.

dgboss: You have a few other options for accessing the property:
1. Object.getOwnPropertyNames
2. Object.keys
3. for const prop in my_object

dgboss (approved):

PR #183, opened 2022-07-22
Add default minZoom, maxZoom values to viewer config

michaelpnelson:
New ESRI basemaps do not have a maxZoom value, and this causes an error to be thrown from the marker-cluster library. This change adds a maxZoom property to map options in viewer-leaflet.js, adds default values for minZoom and maxZoom, and updates viewer configuration docs to add the minZoom and maxZoom properties.


dgboss (approved):
Looks good

PR #185, opened 2022-08-17
Remove ver parameter from geocoder URLs

michaelpnelson:
This URL parameter is no longer used by the Geocoder API.


michaelpnelson:
For more context, see MAP-3266.

dgboss (approved):

PR #187, opened 2022-10-31
Update socket-io.parser for vulnerability

michaelpnelson:
The required version of socket-io.parser is updated to address a critical security vulnerability: https://github.com/advisories/GHSA-qm95-pgcg-qqfq


dgboss (approved):
The change looks good, but as per our discussion with Nicole, this does not warrant a release. The dependency in question is itself a sub-dependency of another npm package which is listed as a 'devDependency' in the smk project's package.json. The vulnerability that is being addressed is never exposed by a hosted SMK application. The socket.io-parser code is only executed when running the project in a development environment.

PR #188, opened 2022-11-07
Update leaflet libs

michaelpnelson:
Update leaflet, esri-leaflet, and leaflet-geoman libraries to latest versions. This doesn't fix the Geomark issue we're investigating, but it's been done and might yield other improvements.


michaelpnelson:
Darren, do these lib updates look good to you? These updates aren't necessary now that the Geomark issue is fixed, but since they've been done, I'd like to include them - they may reduce future issues.

I've done manual testing of these changes plus the recent Geomark changes (testing with other related tools like Identify, Route Planner, and Markup) and it looks good to me.

dgboss (approved):
I think everything is in order. You don't need to make any changes, but in the future consider following a consistent pattern when updating files. For example, esri-leaflet-3.0.2.js is deleted and esri-leaflet-3.0.8.js is added. Compare this to leaflet-1.7.1.css which was updated and renamed to leaflet-1.9.2.css (same for the leaflet geoman css file) instead of doing an add/delete.
-

PR #189, opened 2022-11-08
Fix identify on Esri feature layers

dgboss:
If your SMK app contains an Esri feature layer and you perform an identify, you will see an unhandled error logged to the developer console ("Unknown Geometry Type"). This happens because features returned from `getFeaturesInArea()` in `layer-esri-feature.js` are missing the following attribute: `type: 'Feature'`.

To observe the old behavior:
1. Checkout the smk-clean-bc repo https://github.com/bcgov/smk-clean-bc
2. npm install
3. npm run view
4. Click on a dot on the map and check your browser's dev console


michaelpnelson (approved):
I've confirmed that this change addresses the console error.
I'd be curious to know why setting the type helps.

PR #190, opened 2022-11-16
Geomarks are toggled by pane visibility

michaelpnelson:
Geomarks are currently toggled on and off by adding and removing them from the map. Keeping a full reference to a geomark layer in memory appears to be causing an issue drawing vector basemap layers.

This addresses the issue by storing an id reference to a geomark layer instead of the layer itself. We no longer use the id of the geomark because, when creating a new geomark polygon/multipolygon, we need a reference to bind to the geomark after it's been created. For this purpose, we use a UUID.

Instead of adding and removing geomark layers from the map, we create a custom pane for each geomark layer and add polygon/multipolygons to the custom pane. Toggling then changes the visibility of a custom pane.

This also changes some variable and function names to be clearer.

### DIFF
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -339,7 +339,7 @@ include.module( 'tool-geomark', [
                 },
                 'toggle-geomark': function(idObj) {
                     const geomarkInfo = self.getGeomarkById(idObj.id);
-                    if (!geomarkInfo) {
+                    if (!(geomarkInfo && smk.$viewer.map.getPane(geomarkInfo.layerUUID))) {
                         return;
                     }
                     if (geomarkInfo.isVisible) {
--- a/src/smk/tool/geomark/tool-geomark.js
+++ b/src/smk/tool/geomark/tool-geomark.js
@@ -74,13 +74,20 @@ include.module( 'tool-geomark', [
 
             const CUSTOM_COLOUR = '#ee0077';
 
-            this.createCurrentLayerGroup = function() {
-                const layerGroup = new L.FeatureGroup();
-                smk.$viewer.map.addLayer(layerGroup);
-                return layerGroup;
+            this.createCurrentLayerGroupInfo = function() {
+                const layerUUID = SMK.UTIL.makeUUID();
+                const overlayPane = smk.$viewer.map.getPane('overlayPane');
+                smk.$viewer.map.createPane(layerUUID, overlayPane);
+                const layerGroup = L.featureGroup(null, {
+                    pane: layerUUID
+                }).addTo(smk.$viewer.map);
+                return { 
+                    layerUUID, 
+                    layerGroup 
+                };
             }
 
-            var currentLayerGroup = self.createCurrentLayerGroup();
+            var currentLayerGroupInfo = self.createCurrentLayerGroupInfo();
 
             var currentLayer;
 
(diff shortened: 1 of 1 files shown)


michaelpnelson:
Hi Darren! If you don't have time to look at this in the next few days, no worries, just let me know and I'll find another reviewer. 

I've manually tested these changes by creating and loading polygon and multipolygon geomarks, and tested Geomark tool interaction with the Identify and Markup tools to confirm that previously fixed issues are still working as expected.

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -286,33 +297,38 @@ include.module( 'tool-geomark', [
...
-                                self.updateAndShowAlert('Created geomark: <a href="' + geomarkInfo.url + 
-                                '" target="_new">' + geomarkInfo.url +
-                                '</a>. Save this URL to access your geomark later.');
-                                self.geomarks.push(geomark);
-                                currentLayerGroup = self.createCurrentLayerGroup();
+                        'callback': function(geomarkJson) {
+                            const geomarkInfo = self.toGeomarkInfo(geomarkJson, currentLayerGroupInfo.layerUUID);
+                            if (geomarkInfo) { 
+                                // Layer group must be removed from map before we can reassign its pane, then added back
+                                smk.$viewer.map.removeLayer(currentLayerGroupInfo.layerGroup);
```
dgboss, line 319: Why do you have to reassign the pane with the same value?

src/smk/tool/geomark/tool-geomark.js
```diff
@@ -322,16 +338,16 @@ include.module( 'tool-geomark', [
...
-                    const geomark = self.getGeomarkById(idObj.id);
-                    if (!geomark) {
+                    const geomarkInfo = self.getGeomarkById(idObj.id);
+                    if (!geomarkInfo) {
                         return;
                     }
-                    if (smk.$viewer.map.hasLayer(geomark.drawingLayer)) {
-                        smk.$viewer.map.removeLayer(geomark.drawingLayer);
+                    if (geomarkInfo.isVisible) {
+                        smk.$viewer.map.getPane(geomarkInfo.layerUUID).style.display = 'none';
```
dgboss, line 346: is getPane() guranteed to return an object? Even if the UUID isn't found? If so then this is probably safe. If not, you should add a defensive null check before accessing style.display.

dgboss (approved):
Nice workaround. I've approved but have a couple questions.

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: When creating a geomark, we start by adding polygon layer(s) to the layer group. It seems like the svg element should then get added as a child of the pane's div element, but this doesn't happen, and so without this remove/set-pane/add, the svg stays outside the pane's div. Only by doing all 3 does it seem to be moved (re-added) as a child of the pane div in the DOM. Some more context from a comment in the task I'm tracking this with:

"I ran into a problem with the strategy (of using custom panes). I found this works well when loading a geomark because we create a layer for the polygon(s) and set the custom pane. However this didn’t work as well when creating a geomark, because we create it layer by layer (in case we want a multipolygon) and then add each layer to a FeatureGroup. Unfortunately, Leaflet API favours assigning a layer to a Pane when the layer is created, so it can be added to the DOM without having to worry about moving it later. So a layer effectively cannot be moved from one pane to a different pane.

However I found in [How to change polygon pane in Leaflet (custom pane)](https://stackoverflow.com/questions/42254868/how-to-change-polygon-pane-in-leaflet-custom-pane) that it is possible to remove the layer from the map, change the pane in a setStyle() method, and add it back to the map. This strategy is working for creating a geomark, and so far, I haven’t found any faults after thorough manual testing."

src/smk/tool/geomark/tool-geomark.js
michaelpnelson: Good catch! If the UUID isn't found, it returns undefined. I'll add a check in another commit.

PR #200, opened 2023-01-09
Update version to 1.1.4

michaelpnelson:
The npm publish of 1.1.3 was not correct as I hadn't rebuilt files. An attempt to unpublish was rejected, so I want to increment, re-release, and republish to npm as 1.1.4.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@bcgov/smk",
-  "version": "1.1.3",
+  "version": "1.1.4",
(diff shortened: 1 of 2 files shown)


dgboss:
What do you mean by 're-built files?'

michaelpnelson:
> What do you mean by 're-built files?'

I'd pulled from master but forgot to build to the "deploy" directory, so when I published, recent changes weren't included. An awkward mistake :(

dgboss (approved):

PR #203, opened 2023-01-19
Can hide Esri feature layer legend

michaelpnelson:
Esri feature layers can have multiple legends to reflect differing symbology that is based on differing values for a field. When the "where" option is used to omit features based on such a field value, the legend for the omitted value still appears.

This change introduces a 'hideLegend' parameter to Esri feature layer configuration. When the names of one or more legend values are used, those legends are omitted from viewer display context and do not display in UI legends.

### DIFF
--- a/src/smk/viewer-leaflet/layer/layer-esri-feature-leaflet.js
+++ b/src/smk/viewer-leaflet/layer/layer-esri-feature-leaflet.js
@@ -36,12 +36,14 @@ include.module( 'layer-leaflet.layer-esri-feature-leaflet-js', [ 'layer.layer-es
                 // cfg.drawingInfo.renderer.symbol.url = this.resolveUrl( cfg.drawingInfo.renderer.symbol.url )
                 cfg.drawingInfo.renderer.symbol.url = ( new URL( cfg.drawingInfo.renderer.symbol.url, document.location ) ).toString()
         }
-        
+
         var layer = L.esri.featureLayer( cfg )
         
         if ( layers[ 0 ].legendCacheResolve ) {
+            const hideLegend = layers[0].config.hideLegend ? layers[0].config.hideLegend : "";
             layer.legend( function ( err, leg ) {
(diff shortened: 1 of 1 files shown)


michaelpnelson:
The use case for this is the Clean BC SMK app. The Esri feature layer "Electric Charging Stations in Canada" is configured to only include public charging stations, but we still display both legend values we get for the layer ("public" and "private").

src/smk/viewer-leaflet/layer/layer-esri-feature-leaflet.js
```diff
@@ -36,12 +36,14 @@ include.module( 'layer-leaflet.layer-esri-feature-leaflet-js', [ 'layer.layer-es
...
-        
+
         var layer = L.esri.featureLayer( cfg )
         
         if ( layers[ 0 ].legendCacheResolve ) {
+            const hideLegend = layers[0].config.hideLegend ? layers[0].config.hideLegend : "";
             layer.legend( function ( err, leg ) {
-                layers[ 0 ].legendCacheResolve( err ? null : leg.layers[ 0 ].legend )
+                layers[0].legendCacheResolve(err ? null : 
+                    leg.layers[0].legend.filter(lg => lg.values.every(v => !hideLegend.includes(v))));
```
dgboss: What are the requirements for the hideLegend config item? Is it a value representing a single string? Space delimited strings? Comma delimited strings?

It might be more intuitive from a configuration point of view for hideLegend to be an array of strings. You could them do a case-insensitive comparison of your legend values against each element of hideLegend instead of hideLegend.includes(v).

dgboss (approved):
I think the changes will work, but consider my comment.

src/smk/viewer-leaflet/layer/layer-esri-feature-leaflet.js
michaelpnelson: Good point about case - I'll update for that, and I should be able to work out how to do an array <-> array comparison here.

dgboss (approved):

PR #205, opened 2023-01-30
Improve ordering of layers

michaelpnelson:
This improves the ordering of layers as they display when an SMK app is first loaded. The layer order follows the order of the 'display' section of SMK config for the layers tool EXCEPT GeoJSON/'vector' layers still appear on top.

This adds a 'layers' pane with a z-index of 300 to avoid conflict with other standard layers used by Leaflet. The zIndex assigned from visible layers in viewer.js is corrected to increase on iteration. Code for Leaflet layers is updated to create a new pane for each layer and set each layer's style on the zIndex from view.js.

### DIFF
--- a/src/smk/viewer-leaflet/layer/layer-esri-tiled-leaflet.js
+++ b/src/smk/viewer-leaflet/layer/layer-esri-tiled-leaflet.js
@@ -13,6 +13,11 @@ include.module( 'layer-leaflet.layer-esri-tiled-leaflet-js', [ 'layer.layer-esri
     SMK.TYPE.Layer[ 'esri-tiled' ][ 'leaflet' ].create = function ( layers, zIndex ) {
         if ( layers.length != 1 ) throw new Error( 'only 1 config allowed' )
 
+        var self = this;
+        const layerPane = self.map.getPane(SMK.TYPE.Viewer.leaflet.prototype.layerPane);
+        const thisLayerPane = self.map.createPane(String(layers[0].config.id), layerPane);
+        thisLayerPane.style.zIndex = zIndex;
+
         var serviceUrl  = layers[ 0 ].config.serviceUrl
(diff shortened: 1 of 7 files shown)


michaelpnelson:
I've been testing this with an SMK app with several layers including each Leaflet type, and I'm happy to share its config and files and/or demo it.

michaelpnelson:
The use case for this is the CleanBC app. When loaded, the public charging stations (Esri feature layer) obscure the fleet charging stations layer and buildings layer (GeoJSON), despite the public charging stations being at the bottom of the layer order.

src/smk/viewer-leaflet/layer/layer-esri-tiled-leaflet.js
```diff
@@ -13,6 +13,11 @@ include.module( 'layer-leaflet.layer-esri-tiled-leaflet-js', [ 'layer.layer-esri
     SMK.TYPE.Layer[ 'esri-tiled' ][ 'leaflet' ].create = function ( layers, zIndex ) {
         if ( layers.length != 1 ) throw new Error( 'only 1 config allowed' )
 
+        var self = this;
```
dgboss, line 16: I'm not sure it's 'safe' to mess around with the ordering of tiled layers like this. In general, the tiled layers should appear below all other layers.

dgboss (approved):

src/smk/viewer-leaflet/layer/layer-esri-tiled-leaflet.js
michaelpnelson: Thanks. I think I will update the changeset to remove the 'layers' pane that all new layer panes are added to, and instead add new layer panes to the default pane for each layer type ('tile' and 'overlay', https://leafletjs.com/reference.html#map-pane) to ensure tile layers stay below vector layers.

PR #206, opened 2023-02-08
Update install doc with cmd to launch

michaelpnelson:
Our install page documents how to install SMK and create an SMK app, but not how to launch it. This adds information on how to launch an SMK app.


rustyjux (approved):
Short and sweet but I think it gets the point across!

PR #207, opened 2023-02-09
Vector/GeoJSON layer can be conditionally styled

michaelpnelson:
Hi Darren. This changeset is a bit rough but I wanted to get early feedback on it.

This changeset is in response to CleanBC requests to style shapes differently based on feature values. A specific use case is changing colour based on charging station stall type. I have a WIP branch for this at https://github.com/bcgov/smk-clean-bc/tree/wip_geojson_conditional_styling.

First, please let me know if you know of a simpler way to do this - maybe there's something I'm missing as I don't have your experience with what is done and can be done with web mapping. Or perhaps there's good reason why we shouldn't even consider it? 

I've also changed the way styling is done for this layer type - instead of applying a style to the layer, it's applied functionally to each shape. 

We may want to do this for other layer types - Esri feature layers may be in scope for this - so I anticipate doing further work to see if we can abstract this out for better reuse.

There's also an issue with the legend as it will only show one style at present. Currently it will use values in the "style" element of layer configuration (ignoring overrides in the new "conditionalStyles" element). I may look into updating legend code for this layer.

### DIFF
--- a/src/smk/viewer-leaflet/layer/layer-vector-leaflet.js
+++ b/src/smk/viewer-leaflet/layer/layer-vector-leaflet.js
@@ -143,8 +143,8 @@ include.module( 'layer-leaflet.layer-vector-leaflet-js', [ 'layer.layer-vector-j
                 const overlayPane = self.map.getPane('overlayPane');
                 const layerPane = self.map.createPane(layers[0].config.id, overlayPane);
                 layerPane.style.zIndex = zIndex;
-                
-                var layer = new L.geoJson( null, {
+
+                var layerOptions = {
                     coordsToLatLng: projectCoord,
                     pointToLayer: function ( geojson, latlng ) {
                         // return markerForStyle( self, latlng, layers[ 0 ].config.style )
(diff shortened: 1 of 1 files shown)


src/smk/viewer-leaflet/layer/layer-vector-leaflet.js
```diff
@@ -154,14 +154,36 @@ include.module( 'layer-leaflet.layer-vector-leaflet-js', [ 'layer.layer-vector-j
...
+                        var combinedStyle = Object.assign({}, defaultStyle);
+                        layers[0].config.conditionalStyles.forEach(conditionalStyle => {
+                            if (!Object.keys(feature.properties).includes(conditionalStyle.property)) {
+                                console.debug(`The feature property ${conditionalStyle.property} was not found; conditional styling will not be applied for this property.`);
+                                return convertStyle(combinedStyle, feature.geometry.type);
+                            }
+                            conditionalStyle.conditions.forEach(condition => {
+                                if (condition.value === feature.properties[conditionalStyle.property]) {
+                                    Object.assign(combinedStyle, condition.style);
+                                    return;
```
dgboss: Overall I think this is a very reasonable approach. You're covering a very common use case. The 'Esriesque' term for this would be a unique value renderer. The esri-leaflet package may already have similar functionality out of the box for Esri feature layers.

I could be understanding things incorrectly, but there might be a little bit excess work being done in lines 171-174. What is the purpose of the return statement on line 174? On line 171 you use 'forEach' to iterate through all conditions, but I think you can actually stop once you find a matching condition. If this is the case, you can't 'break out of' a forEach 'loop' as it is a function, not a loop. You could replace the forEach call with a for loop which you can break out of once you find a matching condition, or filter your array of conditions first.

michaelpnelson: Thanks!

I like your suggestion of filtering conditions and will push a change implementing it.

dgboss (approved):

PR #208, opened 2023-02-13
Vector layer docs updated with conditionalStyles

michaelpnelson:
This updates documentation for vector/GeoJson layers to include the new "conditionalStyles" attribute.


rustyjux (approved):
👍 Nice having the example.

PR #210, opened 2023-02-14
Update installation.md

rustyjux:
Tweaks to the 'installation' docs

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -22,7 +22,7 @@ Test that this worked:
 
     > smk help
 
-You should see the help information for `smk-cli`. If you don't and are using Windows, then it's likely your npm folder (`%APPDATA%\npm`) is not part of the system PATH variable. The simplest way to add it is with via the [Control Panel](https://docs.oracle.com/en/database/oracle/machine-learning/oml4r/1.5.1/oread/creating-and-modifying-environment-variables-on-windows.html). After adding the npm folder to path, `smk help` should work.
+You should see the help information for `smk-cli`. If you don't and are using Windows, then it's likely your npm folder (`%APPDATA%\npm`) is not part of the system PATH variable. The simplest way to add it is with via the [Control Panel](https://www.computerhope.com/issues/ch000549.htm). After adding the npm folder to path, `smk help` should work.
 
 ### Creating a project
 Change to the a directory where you keep your projects, and create a new SMK application (change `my-new-app` to whatever you like).
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -6,7 +6,7 @@ There are a variety of ways to install Simple Map Kit (SMK) into your applicatio
 
 ## Use `smk create` to make a new app
 
-If you want to create a new application using SMK, then the easiest way is with the [`smk-cli`](https://github.com/bcgov/smk-cli) (Command Line Interface), a utility for assisting with the creation, modification and development of SMK projects.
+If you want to create a new application using SMK, then the easiest way is with [`smk-cli`](https://github.com/bcgov/smk-cli) (Command Line Interface), a utility for assisting with the creation, modification and development of SMK projects.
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -19,8 +22,9 @@ Test that this worked:
 
     > smk help
 
-You should see the help information for `smk-cli`.
+You should see the help information for `smk-cli`. If you don't and are on Windows, then it's likely your npm folder (`%APPDATA%\npm`) is not part of the system PATH variable. The simplest way to add it is with the [GUI](https://www.architectryan.com/2018/03/17/add-to-the-path-on-windows-10/). After adding the npm folder to path, `> smk help` should work.
```
michaelpnelson: Nit: "on Windows" - maybe "using Windows"? 

For the link - "GUI" is good but I think we could be more specific - maybe "Control Panel"? And I would advise using a more authoritative link as this one is more likely to go away (if the author deletes his blog, say) - I'd recommend a link from Microsoft.

I'd remove the '>' (prompt) with 'smk help' here - I think the prompt works well when the cmd is its own line but looks a bit awkward to me inline like this.

docs/installation.md
```diff
@@ -2,12 +2,15 @@
 
 # Installing SMK
 
-There are a variety of ways to install SMK into your application.
+There are a variety of ways to install Simple Map Kit (SMK) into your application.
 
+## Use `smk create` to make a new app
 
-## Use `smk create`
+If you want to create a new application using SMK, then the easiest way is with the [`smk-cli`](https://github.com/bcgov/smk-cli) (Command Line Interface), a utility for assisting with the creation, modification and development of SMK projects.
```
michaelpnelson: I'd go with "with smk-cli" instead of "with **the** smk-cli", like the way we call SMK "SMK" and not "the SMK".

docs/installation.md
```diff
@@ -29,9 +33,9 @@ Change to the a directory where you keep your projects, and create a new SMK app
 You will be asked some questions about your new application.
 Once they are answered you will have a new skeleton application at `projects/my-new-app`.
 
-You will also be asked whether you want to run the app. You can choose yes to launch the app immediately. 
+You will also be asked whether you want to edit the app. Choosing yes will launch the interactive editor.
```
michaelpnelson: Thanks, this fixes my recent changes - I should have actually run 'smk create' instead of going from memory :(

docs/installation.md
```diff
@@ -29,9 +33,9 @@ Change to the a directory where you keep your projects, and create a new SMK app
...
 
-You can also run the app at a later time by navigating to the app directory and launching it with NPM:
+To run the app locally at any time, navigate to the app directory and call:
```
michaelpnelson: "and call" - maybe "and run"?

michaelpnelson (changes requested):
Thanks for this. It think it's fine overall. I have a few minor suggestions, plus a nit (which for me is a completely ignorable suggestion - "nit" is a term I picked up in previous work).

docs/installation.md
michaelpnelson: One other thing - "interactive editor" - this launches the web GUI, and maybe that's what you meant, but the CLI wizard is also arguably an interactive editor, so consider making this more specific.

docs/installation.md
rustyjux: Sure.

docs/installation.md
rustyjux: Found a slightly more official page in Oracle docs. Nothing easily found on MS oddly.

docs/installation.md
rustyjux: Changed to 'web GUI editor'

docs/installation.md
michaelpnelson: What do you think of this link instead? https://www.computerhope.com/issues/ch000549.htm The Oracle link has content specific to installing its tool; this is more generic.

rustyjux: Works for me - doesn't have as many images but heck they been around since 1998

michaelpnelson (approved):

PR #213, opened 2023-03-03
Improvements to GeoJSON layer conditional styling

michaelpnelson:
This introduces operators to the conditional styling syntax to allow comparisons beyond equals. Equals is used as the default to maintain backwards compatibility.

Documentation for vector/GeoJSON layer configuration is also updated.

### DIFF
--- a/src/smk/viewer-leaflet/layer/layer-vector-leaflet.js
+++ b/src/smk/viewer-leaflet/layer/layer-vector-leaflet.js
@@ -168,7 +168,7 @@ include.module( 'layer-leaflet.layer-vector-leaflet-js', [ 'layer.layer-vector-j
                                 console.debug(`The feature property ${conditionalStyle.property} was not found; conditional styling will not be applied for this property.`);
                                 return convertStyle(combinedStyle, feature.geometry.type, layerPaneId);
                             }
-                            conditionalStyle.conditions.filter(condition => condition.value === feature.properties[conditionalStyle.property]).forEach(condition => {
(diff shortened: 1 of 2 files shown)


michaelpnelson:
I also looked into ways to allow adding a styling function - an escaped string containing a JavaScript function - to SMK configuration JSON for full flexibility. I was reminded JSON doesn't support functions, and using the Function constructor causes a jshint warning as the Function constructor is an eval, and using an eval sounds like a bad idea.

src/smk/viewer-leaflet/layer/layer-vector-leaflet.js
```diff
@@ -305,6 +305,19 @@ include.module( 'layer-leaflet.layer-vector-leaflet-js', [ 'layer.layer-vector-j
             }
     }
 
+    function matches(value, condition) {
+        switch(condition.operator) {
+            case '>': return !Number.isNaN(value) && value > condition.value;
```
dgboss: You might want to consider throwing an error or somehow handle the situation where the value can't be parsed as a number. In the current state, these numerical comparisons would always return false and someone might be left scratching their heads wondering what's going wrong. Even a console.warn() message could be helpful.

dgboss (approved):

src/smk/viewer-leaflet/layer/layer-vector-leaflet.js
michaelpnelson: Thanks for this suggestion. I'll log a message when the value in a numerical comparison is not a number.

PR #214, opened 2023-03-06
Increment version to 1.1.7

michaelpnelson:


rustyjux (approved):
Not much to reviewing version increments!

michaelpnelson:
FYI I won't merge this right away. I might be able to merge one or two other features before we merge this and do a release. But it's great to have this approved now so that I can release when I need to.

rustyjux:
Now that seems like a sneaky way to slip in features without my careful review!

PR #215, opened 2023-03-07
Add range and marker to Directions tool

michaelpnelson:
This adds a UI control to direction options to allow input of a range as a positive integer. When a range value exists, the route planner now captures the point along the route where the range limit is met. The directions tool shows a new marker at the range limit point.


src/smk/api/route-planner.js
```diff
@@ -161,6 +160,16 @@ include.module( 'api.route-planner-js', [ 'jquery', 'util' ], function () {
...
+                if (option.rangeKm > 0 && data.distance && option.rangeKm < data.distance) {
+                    var distanceKm = 0;
+                    for (var i = 0; i < data.route.length + 1; i += 1) {
+                        distanceKm += turf.distance(data.route[i], data.route[i+1], {units: "kilometers"});
+                        if (distanceKm >= option.rangeKm) {
+                            data.rangeLimit = data.route[i];
+                            break;
+                        }
+                    }
+                }
```
rustyjux, line 172: I like. Seems like performance is fair (~2 s), even for long routes with longer ranges (750 km)

src/smk/tool/directions/panel-directions-options.html
```diff
@@ -8,6 +8,8 @@
     <template slot="commands">
     </template>
 
+    <enter-input v-bind:value="rangeKm" v-on:change="!busy && $$emit('change', {rangeKm: $event});">Range (km)</enter-input>
```
rustyjux, line 11: We had talked about a slider before, but entering a numerical value is quick and easy to be precise. Though slider would maybe be preferable on mobile.

rustyjux:
Looks good overall.

I like how modifying range updates the range marker without needing to run the route again.

Marker dropshadow looks nice on the side it is on, but the marker white border doesn't have great contrast against the (current) light coloured basemap elsewhere.

rustyjux (approved):

PR #216, opened 2023-03-08
Fix error in vector docs

michaelpnelson:


rustyjux (approved):
does this really need a review? :P

michaelpnelson:
This is trivial, yes, but for this repo, being public, we have branch protection rules requiring at least 1 reviewer for pull requests. And it shares a wee bit of knowledge.

PR #219, opened 2023-03-17
Add null check when validating numbers

michaelpnelson:
When applying a numerical comparison to a feature value that can be null, null values trigger a console warning on null values instead of just non-number values. This can result in hundreds of extra console warnings.


rustyjux (approved):
Good catch, simple solution.

PR #220, opened 2023-03-17
Vector layer legend includes conditional styling

michaelpnelson:
This updates legend code to include conditional styling.

* A "Label" attribute is added to conditional styling configuration for use in legends
* The documentation for vector layers is updated to include "label"
* Legend code includes conditional styling, indented below the layer name and its default style image

### DIFF
--- a/src/smk/layer/layer-vector.js
+++ b/src/smk/layer/layer-vector.js
@@ -21,54 +21,82 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
         var cv = $( '<canvas width="' + width * mult + '" height="' + height + '">' ).get( 0 )
         var ctx = cv.getContext( '2d' )
 
-        var styles = [].concat( self.config.style )
+        const legendData = [];
+        legendData.push({
+            title: self.config.legend.title || self.config.title,
+            style: self.config.style
+        });
+
+        if (self.config.conditionalStyles) {
+            self.config.conditionalStyles.forEach(cs => {
+                const defaultStyle = self.config.style ? self.config.style : {};
+                self.config.conditionalStyles.forEach(conditionalStyle => {
+                    conditionalStyle.conditions.forEach(condition => {
+                        const combinedStyle = Object.assign({}, defaultStyle);
+                        Object.assign(combinedStyle, condition.style);
+                        legendData.push({
+                            title: condition.label || condition.value,
+                            style: combinedStyle,
+                            indent: true
+                        });
+                    });
+                });
+            });
+        }
 
         return SMK.UTIL.resolved( 0 )
             .then( drawPoint )
             .then( drawLine )
             .then( drawFill )
             .then( function () {
-                return [ {
-                    url: cv.toDataURL( 'image/png' ),
-                    title: self.config.legend.title || self.config.title
-                } ]
+                return legendData;
             } )
 
         function drawPoint( offset ) {
             if ( !self.config.legend.point ) return offset 
 
-            return SMK.UTIL.makePromise( function ( res, rej ) {
-                if ( styles[ 0 ].markerUrl ) {
-                    var img = $( '<img>' )
-                        .on( 'load', function () {
-                            var r = img.width / img.height
-                            if ( r > 1 ) r = 1 / r
-                            ctx.drawImage( img, offset, 0, height * r, height )
-                            res( offset + width )
-                        } )
-                        .on( 'error', res )
-                        .attr( 'src', viewer.resolveAttachmentUrl( styles[ 0 ].markerUrl, null, 'png' ) )
-                        .get( 0 )
-                }
-                else {
+            if (legendData.length === 1 && legendData[0].style.markerUrl) {
+                loadMarkerImage(offset)
+                .then((offset) => {
+                    legendData[0].url = cv.toDataURL( 'image/png' );
+                    return offset;
+                });
+            } else {
+                legendData.forEach(leg => {
+                    const st = leg.style;
                     ctx.beginPath()
-                    ctx.arc( offset + width / 2, height / 2, styles[ 0 ].strokeWidth / 2, 0, 2 * Math.PI )
+                    ctx.arc( offset + width / 2, height / 2, st.strokeWidth / 2, 0, 2 * Math.PI )
                     ctx.lineWidth = 2
-                    ctx.strokeStyle = cssColorAsRGBA( styles[ 0 ].strokeColor, styles[ 0 ].strokeOpacity )
-                    ctx.fillStyle = cssColorAsRGBA( styles[ 0 ].fillColor, styles[ 0 ].fillOpacity )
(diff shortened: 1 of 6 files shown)


src/smk/layer/layer-vector.js
```diff
@@ -21,54 +21,82 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
         var cv = $( '<canvas width="' + width * mult + '" height="' + height + '">' ).get( 0 )
         var ctx = cv.getContext( '2d' )
 
-        var styles = [].concat( self.config.style )
+        const legendData = [];
+        legendData.push({
+            title: self.config.legend.title || self.config.title,
```
dgboss: I'm not sure if `legend` is required configuration, but you might want some defensive coding here to protect against trying to access the `title` property on something undefined.
eg. `title: (self.config.legend && self.config.legend.title) || self.config.title,`

src/smk/layer/layer-vector.js
```diff
@@ -21,54 +21,82 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
...
+            style: self.config.style
+        });
+
+        if (self.config.conditionalStyles) {
+            self.config.conditionalStyles.forEach(cs => {
```
dgboss: I haven't tried running this code, but it seems odd to have `self.config.conditionalStyles.foreach()` nested within another  `self.config.conditionalStyles.foreach()` ie line 33 nested within 31.

src/smk/layer/layer-vector.js
```diff
@@ -21,54 +21,82 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
...
-                            if ( r > 1 ) r = 1 / r
-                            ctx.drawImage( img, offset, 0, height * r, height )
-                            res( offset + width )
-                        } )
-                        .on( 'error', res )
-                        .attr( 'src', viewer.resolveAttachmentUrl( styles[ 0 ].markerUrl, null, 'png' ) )
-                        .get( 0 )
-                }
-                else {
+            if (legendData.length === 1 && legendData[0].style.markerUrl) {
```
dgboss, line 56: Is `legendData[0]` guaranteed to have a `style` property?

src/smk/layer/layer-vector.js
```diff
@@ -90,14 +119,16 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
         function drawFill( offset ) {
             if ( !self.config.legend.fill ) return offset 
 
-            styles.forEach( function ( st ) {
+            legendData.forEach( function ( leg ) {
+                const st = leg.style;
                 // var w = self.config.style.strokeWidth
```
dgboss: I'd suggest cleaning up unnecessary (commented out) code.

src/smk/layer/layer-vector.js
```diff
@@ -21,54 +21,82 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
...
-                        .on( 'error', res )
-                        .attr( 'src', viewer.resolveAttachmentUrl( styles[ 0 ].markerUrl, null, 'png' ) )
-                        .get( 0 )
-                }
-                else {
+            if (legendData.length === 1 && legendData[0].style.markerUrl) {
+                loadMarkerImage(offset)
+                .then((offset) => {
+                    legendData[0].url = cv.toDataURL( 'image/png' );
+                    return offset;
```
dgboss, line 60: Do you know what the purpose of returning `offset` is? Is it necessary?

src/smk/layer/layer-vector.js
```diff
@@ -21,54 +21,82 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
...
+                });
+            } else {
+                legendData.forEach(leg => {
+                    const st = leg.style;
```
dgboss: Point of personal preference, but variable names that match the intended meaning make code more readable for others. eg `const legendStyle = legendItem.style;`

No need to change, just a thought for future consideration.

src/smk/tool/layers/layer-display.html
```diff
@@ -37,6 +37,7 @@
             <div class="smk-legend-item"
                 v-for="legend in display.legends"
             >
+                <span v-if="legend.indent">&nbsp;&nbsp;</span>
```
dgboss: Can this be accomplished with CSS?

src/smk/tool/legend/legend-display.html
```diff
@@ -22,6 +22,7 @@
             <div class="smk-legend-item" 
                 v-for="legend in display.legends"
             >
+                <span v-if="legend.indent">&nbsp;&nbsp;</span>
```
dgboss: Can this be accomplished with CSS?

dgboss (changes requested):
My only real concern is the double-nested foreach on the same object.

src/smk/layer/layer-vector.js
michaelpnelson: Probably a copy/paste error ... this change went through a lot of iterations before I pushed it to review. I'll update so there's only one.

src/smk/layer/layer-vector.js
michaelpnelson: Done.

src/smk/layer/layer-vector.js
michaelpnelson: Yes - this is set earlier in line 27 using the layer's config.style.

src/smk/layer/layer-vector.js
michaelpnelson: Done.

src/smk/layer/layer-vector.js
michaelpnelson: I don't believe it's used. We do need to have a return value in promises though, so I left it all as-is for consistency, and the chance this is used in some other context.

dgboss: Is the return value from a promise a requirement of the SMK framework? It's not required by JavaScript.

michaelpnelson: I might be misinterpreting, but I got the impression a return is required from reading reference like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises - "Important: Always return results, otherwise callbacks won't catch the result of a previous promise (with arrow functions, () => x is short for () => { return x; }). If the previous handler started a promise but did not return it, there's no way to track its settlement anymore, and the promise is said to be "floating".".

src/smk/layer/layer-vector.js
michaelpnelson: I went ahead and renamed these variables for clarity.

src/smk/tool/layers/layer-display.html
michaelpnelson: Yes (here in layer as well as in legend). Good point; there's no semantic meaning in a nonbreaking space, which might manifest in text-to-speech readers or other alternate tools. I've created styles and applied them.

dgboss (approved):
Looks good!

PR #222, opened 2023-03-27
Vector layer can hide layer symbology when conditional styling used

michaelpnelson:
In some uses of conditional styling, styling is applied to feature values in a way that each feature is included and matches one style, similar to a pie chart. In such situations, including a style for the entire layer in a legend is unnecessary.

This adds a "hideLayer" property to legend configuration in vector layers and to legend display HTML. When added to a vector/GeoJSON layer's legend configuration and set to true, the symbology for the layer (but not its conditional styling) is omitted in the legends.

### DIFF
--- a/docs/configuration/layers/vector.md
+++ b/docs/configuration/layers/vector.md
@@ -145,9 +145,9 @@ Optional attributes to configure the display of the layer in legends.
 
 Set the shape of the legend swatch. "line" is the default. 
 
-`"includeOtherLegendWithDefaultStyling": Boolean`
+`"excludeOtherLegendWithDefaultStyling": Boolean`
 
-When <a href="#conditionalstyles-property">conditional styles</a> are used, set this to true to add an "Other" legend with a swatch using the styling in the <a href="#style-property">style</a> to represent features that are not matched by any of the conditions defined in conditional styling.
+By default, when <a href="#conditionalstyles-property">conditional styles</a> are used, an "Other" legend will be added with a swatch using the styling in the <a href="#style-property">style</a> to represent features that are not matched by any of the conditions. Set `excludeOtherLegendWithDefaultStyling` to true to exclude the "Other" legend for situations where all of a layer's features will be matched by the conditions and an "Other" legend is not necessary.
 
 ## DataUrl Property
 `"dataUrl": String`
--- a/docs/configuration/layers/vector.md
+++ b/docs/configuration/layers/vector.md
@@ -23,7 +23,8 @@ Click on a property name for more information:
     <a href="#clusteroption-property"       >"clusterOption"</a>:       null,
     <a href="#useheatmap-property"          >"useHeatmap"</a>:          false,
     <a href="#style-property"               >"style"</a>:               null,
-    <a href="#conditionalstyles-property"  >"conditionalStyles"</a>:   null,
+    <a href="#conditionalstyles-property"   >"conditionalStyles"</a>:   null,
+    <a href="#legend-property"              >"legend"</a>:              null,
     <a href="#dataUrl-property"             >"dataUrl"</a>:             null
 } ] }
 </pre>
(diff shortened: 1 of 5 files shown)


dgboss:
Do you have an example of this behavior or a description on how to reproduce?

michaelpnelson:
> Do you have an example of this behavior or a description on how to reproduce?

I think it's easiest to show the difference :) 

Our use case is for the CleanBC app. The Ministry EVCS layer uses conditional styling to set the colour of a feature based on the stall type. Each feature is one of the two stall types we conditionally style for, but we still show a style for the layer itself, which seems redundant:

Screenshot 2023-03-29 at 8 33 41 AM

With this change, and a change to CleanBC to add "hideLayer": true to the layer's legend config, we modify the legend display to make more sense:

Screenshot 2023-03-29 at 8 36 05 AM

dgboss:
Thanks for the explanation and pictures! My recommendation would be to never show the swatch for the layer when conditional styling is present. If you want to be able to show the default style for the layer, it is common practice to display a 'default' entry below the entries for conditional styles. You could have config to show/hide this default symbol and a 'label' property or something like that so you can configure what text to display instead of 'default'. This is a more typical approach for web maps.
image

Also, the layer name needs some whitespace to the left.

michaelpnelson:
> Thanks for the explanation and pictures! My recommendation would be to never show the swatch for the layer when conditional styling is present. If you want to be able to show the default style for the layer, it is common practice to display a 'default' entry below the entries for conditional styles. You could have config to show/hide this default symbol and a 'label' property or something like that so you can configure what text to display instead of 'default'. This is a more typical approach for web maps. image
> 
> Also, the layer name needs some whitespace to the left.

Thanks for this suggestion. I like the idea of making things more consistent by always omitting legend styling for a layer's name when the layer has conditional styling. 

Do you happen to have a link or two handy where this style of legend appears? I don't have much reference for this style.

I think I'd like to get more feedback from others (especially UX advice from Jeffrey) about this approach also. I agree that it would be good to be able to configure "default" labeling.

dgboss:
> 

The following link has a legend with an 'Other' entry which is analogous to the 'default'.
https://developers.arcgis.com/javascript/latest/sample-code/visualization-location-types/

michaelpnelson:
Here is what legends look like after changes in the latest commit. The new "includeOtherLegendWithDefaultStyling" legend property has been set to true for layer "Electric Charging Stations in BC".

Screenshot 2023-03-30 at 10 20 52 AM

src/smk/layer/layer-vector.js
```diff
@@ -24,7 +24,8 @@ include.module( 'layer.layer-vector-js', [ 'layer.layer-js' ], function () {
         const legendData = [];
         legendData.push({
             title: self.config.legend && self.config.legend.title || self.config.title,
-            style: self.config.style
+            style: self.config.style,
+            hasConditionalStyling: self.config.conditionalStyles || false
```
dgboss: While JavaScript allows the assignment of an object or a boolean to the same variable, it is not good practice. I'd suggest assigning true/false or assigning the object/undefined to `hasConditionalStyling`.

dgboss: My choice would be true/false as you never actually access the object properties.

dgboss (approved):

src/smk/layer/layer-vector.js
michaelpnelson: I'm checking for the existence of the "conditionalStyles" property in config, but yes, it is kind of dumb to actually assign it when it's present :) . I'll update this to do a better check of existence.

docs/configuration/layers/vector.md
```diff
@@ -136,6 +136,19 @@ This sample configuration styles features having a `Charging_Level` value of 3 o
...
+## Legend Property
+`"legend": Object`
+
+Optional attributes to configure the display of the layer in legends.
+
+`"point", "line", "fill": Boolean`
+
+Set the shape of the legend swatch. "line" is the default. 
+
+`"includeOtherLegendWithDefaultStyling": Boolean`
```
michaelpnelson: On further thought, I think I will reverse this to something like "hideOtherLegendWithDefaultStyling". I think we should include the default styling as a ... default :) . I think most uses of conditional styling will want to include the default, and I want to favour inclusion.

PR #226, opened 2023-04-28
Can override Other legend label

michaelpnelson:
This adds an optional configuration parameter to vector/GeoJSON layers to replace "Other" as a legend label when conditional styling is used.

Documentation for the vector layer type is also updated.

This also increments SMK's version.


rustyjux (approved):
Great feature to have (alongside excluding other from the legend).
Looks like a clean and simple implementation. I haven't deployed locally but shouldn't be too much to test on this one.

dgboss (approved):

PR #236, opened 2023-10-20
Replace StamenTonerLight basemap with BC Gov basemaps

michaelpnelson:
Hi Darren! This change is to address an issue with Stamen tiles. Free (non-API-key) support is ending at the end of October. This change replaces StamenTonerLight with two basemaps from GeoBC (BC Basemap without hillshade, BC Basemap with hillshade) (https://governmentofbc.maps.arcgis.com/home/search.html?restrict=true&sortField=relevance&sortOrder=desc&searchTerm=tags%3A%22GeoBC%22&focus=layers-weblayers-tiles#content).

There may be further tweaks to the names (still integrating feedback from others), so this isn't ready for merging yet, but I think the functionality is ready for review.

I'll invite you to a corresponding review in smk-cli as well.

Thanks for your eyes!


michaelpnelson:
I missed a change ... the debug apps have been updated to use an earlier name "GeoBC" - I will push a commit using the "BCGov" name.

dgboss (approved):
Things seem to be working for the most part but I have a few comments/suggestions:

1. Change the back-up/default option for the minimap tool from Topographic to one of the BCGov base maps in case no esriApi key is configured.
2. If you enable the basemap control and don't have an esriApi key configured you get an error about the missing key. Since there is now more than one free option, consider showing the free basemaps when no key is present and show all basemaps when the key is configured.
3. It might be nice to show the BCGov basemap by default when the baseMap config option is missing from the view section of smk-config.json. Right now the behaviour is to complain about a missing esriApi key. This would allow you to remove the baseMap config option from all the debug samples!

michaelpnelson:
Thanks Darren! I appreciate your suggestions. Here's what I did or didn't do for each ...

1 - Using either of the BCGov basemaps doesn't work well for a minimap as the basemap data doesn't go much beyond BC. Here's what it looks like:
Screenshot 2023-10-24 at 8 27 51 AM
I'll think on a better solution because it would be great to not have that popup when the esriApiKey is missing and the Minimap tool is selected, but that will be a separate PR.
2 - I addressed this in the new commit.
3 - This was addressed earlier in the changes to viewer-leaflet.js, lines 229-232:
        // If basemap is not recognized, apply the default
        if(!this.basemap[basemapId]) {
            basemapId = "BCGov";
        }

PR #261, opened 2026-03-05
Update dependency versions to address vulnerabilities

michaelpnelson:
This addresses vulnerabilities found from "npm audit":
16 vulnerabilities (1 low, 1 moderate, 14 high)


michaelpnelson:
Closing this temporarily - testing "npm run" tools shows a bug was introduced by these changes. I'll be adding another commit.

sahejmundi (approved):

michaelpnelson:
These changes don't address vulnerabilities from library "minimatch". Typically I'd address this by adding an override in the "overrides" section to a patched version - for example: 

  "overrides": {
    ...
    "minimatch": ">=3.1.4"
  }

(Version taken from a vulnerability report, https://github.com/advisories/GHSA-23c5-xmqv-rm74.) 

However, when I do this, it breaks functionality. Most of the grunt tasks called by "npm run" stop working. 

I don't think this is a concern though. All of the dependencies in SMK's package.json are "devDependencies" and are only used when we run grunt tasks - which we or a user will do on their workstation. If minimatch were used in the "dependencies" and run on a server, that would be concerning.

ccossette:
Did you try all available versions of minimatch to check for a compatible version that solves the vulnerability?

michaelpnelson:
> Did you try all available versions of minimatch to check for a compatible version that solves the vulnerability?

There are 8 version ranges in the vulnerability report https://github.com/advisories/GHSA-23c5-xmqv-rm74, so I didn't test them all, but I did test a few including the oldest and newest. I also looked at the dependency tree and all dependent versions were in the oldest range (<3.1.4).

